PR #1, opened 2019-11-11
Remove explicitly defined copy operator for style::internal::Font

mymedia2:
It was equivalent of default copy operator generated by compiler itself.
Removing this copy operation, we do not change behavior of the code but
resolve the -Wdeprecated-copy warning from GCC.


john-preston:
Thanks.

PR #7, opened 2020-04-13
Better algorithm for font choosing

ilya-fedin:


ui/style/style_core_font.cpp
```diff
@@ -168,7 +200,18 @@ void StartFonts() {
 		//QFont::insertSubstitution(name, fallback);
 #endif // Q_OS_WIN
 	}
-#endif // !DESKTOP_APP_USE_PACKAGED_FONTS
+
+#ifdef Q_OS_WIN
+	auto list = QStringList();
+	list.append("Microsoft YaHei");
```
NekoInverter, line 206: Why it's Microsoft Yahei instead of Microsoft Yahei UI?

ilya-fedin: I took a list of fonts from FontMod readme

Coxxs:
AFAIK, every Windows system has `Microsoft YaHei`, `Microsoft JhengHei UI`, and `Yu Gothic UI` font. We should change the font priority based on the system region. 

For example, for Traditional Chinese, `Microsoft JhengHei UI` should be prioritized, and for Japanese, `Yu Gothic UI` should be prioritized. `Microsoft YaHei` is designed _just_ for Simplified Chinese.

ilya-fedin:
> We should change the font priority based on the system region.

Qt doesn't prioritize them in any locale. I.e. if these substitutions will be only for japanese/chinese/korean locale, users with english locale will get the old fonts from winxp, as it was before substiutions.

Coxxs:
The issue here is system with Japanese / Tradition Chinese language will use Microsoft YaHei (not `Yu Gothic UI` or `Microsoft JhengHei UI`).

ysc3839:
@ilya-fedin 
> Qt doesn't prioritize them in any locale.

No, Qt has 5 lists and it's selected based on `GetUserDefaultLangID`:
https://github.com/qt/qtbase/blob/44e8c90ad4ae5a3d08651a61ac9f0937fb49d26b/src/platformsupport/fontdatabases/windows/qwindowsfontdatabasebase.cpp#L895-L986

The best solution is to use system "MessageFont", it's implemented but disabled for Qt 5:
https://github.com/qt/qtbase/blob/44e8c90ad4ae5a3d08651a61ac9f0937fb49d26b/src/platformsupport/fontdatabases/windows/qwindowsfontdatabasebase.cpp#L761-L784

ilya-fedin:
> No, Qt has 5 lists and it's selected based on `GetUserDefaultLangID`:

According to your list, it is old winxp fonts, as I said:
```
    "MS UI Gothic",
    "Gulim",
    "SimSun",
    "PMingLiU",
```

ilya-fedin:
btw @ysc3839 thank you for the link, now we can patch Qt instead of substituions, I sent it to @john-preston

Coxxs:
What Qt does is not perfect, but it's _right_. Use Simplified Chinese font for Japanese is not right. 

Telegram 2.0.1 (Japanese System):
image

Telegram 2.1 (Japanese System):
image

You can see the wrong glyph displayed in Telegram 2.1 (Japanese System).

I'm not familiar with Qt, but I guess @ysc3839 means you can use Qt's method (call WinApi SystemParametersInfo, https://github.com/qt/qtbase/blob/44e8c90ad4ae5a3d08651a61ac9f0937fb49d26b/src/platformsupport/fontdatabases/windows/qwindowsfontdatabasebase.cpp#L767) to get default GUI font, and set it as the fallback font.

ilya-fedin:
> call WinApi SystemParametersInfo

> based on `GetUserDefaultLangID`

:thinking:

PR #11, opened 2020-05-04
Update CMakeLists.txt for Clang

lionello:
Workaround for error: `Objective-C 1 was disabled in PCH file but is currently enabled`,
similar to https://github.com/desktop-app/lib_spellcheck/pull/8

This happens because Clang cannot use a PCH emitted while compiling C/C++ for compiling Objective-C sources.

The `-fobjc-weak` was added to suppress the error `Cannot create __weak reference in file using manual reference counting`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,15 @@ generate_emoji(lib_ui emoji_suggestions/emoji_autocomplete.json)
 
 set_target_properties(lib_ui PROPERTIES AUTOMOC ON AUTORCC ON)
 
+# Workaround for error: Objective-C 1 was disabled in PCH file but is currently enabled
+if (CMAKE_CXX_COMPILER_ID MATCHES Clang)
+    file(GLOB ObjcFiles ui/platform/mac/*.mm)
+    set_source_files_properties(${ObjcFiles} PROPERTIES
+        SKIP_PRECOMPILE_HEADERS ON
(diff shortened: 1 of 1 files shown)


john-preston:
:hmm: Why is it not needed for Apple Clang? And it builds the Objective-C++ files while using PCH, not ignoring it (.mm is Objective-C++, not Objective-C).

john-preston:
If you're using non-Xcode CMake generator on macOS I'm afraid this could be a CMake bug. Because in Xcode generator there is three PCH compilations when building Telegram target:

image

First one is with `c++-header` argument, second with `objective-c++-header` argument and third with `c-header` argument. And it uses correct PCH with each file extension while compiling them (.cpp / .c / .mm).

lionello:
Interesting. I was building from command line using Clang from Nixpkgs https://nixos.org/nix 
```
$ clang++ -v
clang version 5.0.2 (tags/RELEASE_502/final)
Target: x86_64-apple-darwin19.4.0
Thread model: posix
InstalledDir: /nix/store/2abb3svdazmy49xhdlqq4wxh4cawjkv5-clang-5.0.2/bin
```

lionello:
I think you're right: it might be a Cmake bug. It's only creating a single PCH.

lionello:
This is a similar issue as what I was having: https://musescore.org/en/node/281503

john-preston:
The fixes you propose should not break usage of PCH in case an Xcode generator is used. Maybe it is better to file a bug report to CMake for other generators.

PR #12, opened 2020-05-04
Patch semibold font as bold to fallback to bold fonts for CJK languages

ilya-fedin:
Fixes https://github.com/telegramdesktop/tdesktop/issues/7775


john-preston:
Thanks!

PR #13, opened 2020-05-08
Add Vazir font for Persian glyphs

ilya-fedin:


john-preston:
Thanks!

ui/style/style_core_font.cpp
```diff
@@ -300,11 +314,9 @@ FontData::FontData(int size, uint32 flags, int family, Font *other)
 	f.setUnderline(_flags & FontUnderline);
 	f.setStrikeOut(_flags & FontStrikeOut);
 
-#ifdef DESKTOP_APP_USE_PACKAGED_FONTS
 	if (_flags & FontSemibold) {
-		f.setWeight(QFont::DemiBold);
+		f.setStyleName("Semibold");
```
hexchain, line 318: May I ask why this is required? It breaks styles for fallback fonts on Linux again. Looks like Qt does not translate `styleName` to style flags, so fonts without a "Semibold" style will not be used regardless.

ilya-fedin: They shouldn't be used

ilya-fedin: Now bold-styled messages should be bold, but semibold UI elements that rendered throught font which doesn't have semibold should be regular, on all platforms

ilya-fedin: Also, open sans should be always when tdesktop is installed, custom fonts aren't supported

ilya-fedin: Anyway, a new release should be soon with semibold style name as a priority and bold style as a fallback
https://github.com/desktop-app/lib_ui/tree/bold

PR #21, opened 2020-09-23
do not treat non-breaking spaces (U+00A0, dec 160) as normal spaces

Mithgol:
This should be an easy (i.e. one-liner) fix for telegramdesktop/tdesktop#6393 unless it would create some undesired side effect [in `bool CutPart(TextWithEntities &sending, TextWithEntities &left, int32 limit)`](https://github.com/desktop-app/lib_ui/blob/d71d2121b1ffc10d2c5565d911bfa31a505d343a/ui/text/text_entity.cpp#L1467) and thus would become not so easy.

### DIFF
--- a/ui/text/text.cpp
+++ b/ui/text/text.cpp
@@ -3408,7 +3408,7 @@ bool IsNewline(QChar ch) {
 }
 
 bool IsSpace(QChar ch, bool rich) {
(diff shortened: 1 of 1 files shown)


Mithgol:
Closed because undesired side effects are possible; rewritten as #22.

PR #42, opened 2021-03-08
Update Vazir font to 27.2.2

MuratKaribay:
Please update this font for Telegram TDesktop
@john-preston


john-preston:
I've updated, thanks.

PR #82, opened 2022-03-01
Less Q_OBJECTs

ilya-fedin:


ui/widgets/input_fields.h
```diff
@@ -56,7 +56,6 @@ enum class InputSubmitSettings {
 };
 
 class FlatInput : public RpWidgetBase<QLineEdit> {
-	// The Q_OBJECT meta info is used for qobject_cast!
```
23rd, line 59: Why it should be removed?

ilya-fedin: Becaue qobject_cast is not used anymors

PR #253, opened 2024-12-29
InputField: only soften line breaks inside format tags

### LINKED ISSUE #28805: Message edit: triple click on a word should select the line, not the whole message
### Steps to reproduce

1. Start editing a message that contains multiple lines
2. In the edit area, triple click a word

### Expected behaviour

The line that contains the word is selected (just like it happens when I triple click a word when writing a new message or when I triple click a word in a message in the chat view)

### Actual behaviour

The whole message is selected

### Operating system

Ubuntu 22.04, i3 version 4.20.1   //   Arch Linux, i3 version 4.24

### Version of Telegram Desktop

5.9   //   5.9.2 beta

### Installation source

Static binary from official website

### Crash ID

_No response_

### Logs

_No response_

kolayne:
In `InputField::PrepareForInsert`, avoid replacing all line break characters with `kSoftLine` (aka `QChar::LineSeparator`) because that makes the selection behavior inconsistent (different for typed text and restored text, e.g., from a draft). Instead, only soften line breaks inside formatting tags.

Should fix telegramdesktop/tdesktop#28805

### DIFF
--- a/ui/widgets/fields/input_field.cpp
+++ b/ui/widgets/fields/input_field.cpp
@@ -1069,11 +1069,6 @@ void ApplyTagFormat(QTextCharFormat &to, const QTextCharFormat &from) {
 			text[position] = ch;
 		}
 	};
-	for (auto i = 0, length = int(data.text.size()); i != length; ++i) {
-		if (newline(i)) {
-			force(i, kSoftLine);
-		}
-	}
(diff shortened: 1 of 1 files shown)


SoeAung95 (approved):

kolayne:
@john-preston, is there any news on this?

PR #254, opened 2025-01-01
`Ui::Animations::Simple`: ensure `.start()` starts from `from`

kolayne:
Unlike the `.change()` method, which configures a running animation to
go to a different state, the `.start()` method should first set the
initial opacity value, then animate.

It is fixed by making sure the private `.prepare()` method always sets
`from`, rather than ignoring it when the animation is already running.

Part of a fix of telegramdesktop/tdesktop#28811.

### DIFF
--- a/ui/effects/animations.h
+++ b/ui/effects/animations.h
@@ -381,16 +381,19 @@ inline void Simple::change(
 		anim::transition transition) {
 	Expects(_data != nullptr);
 
-	prepare(0. /* ignored */, duration);
+	prepare(_data->value /* keep the old value */, duration);
 	startPrepared(to, duration, transition);
 }
 
 inline void Simple::prepare(float64 from, crl::time duration) {
 	const auto isLong = (duration > kLongAnimationDuration);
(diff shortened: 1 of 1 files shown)


ui/effects/animations.h
```diff
@@ -69,6 +69,7 @@ class Simple final {
 		anim::transition transition = anim::linear);
 	void stop();
 	[[nodiscard]] bool animating() const;
+	[[nodiscard]] bool fadingOut() const;
 	[[nodiscard]] float64 value(float64 final) const;
```
kolayne, line 73: With `.fadingOut()`, there's more information that a user of the class can extract, and it's sufficient for the intended use, but there still could be other uses in the future that would like even more information.

Should we introduce a more generic API (e.g., allow users to query the values of `from` and `delta`, like they can now only do with the current `value`)? Or would you rather that we export as little as the current use of the class requires?

kolayne:
Closing, as it's no longer a part of the original fix (although I feel like it is still the right behavior considering the method names).

PR #262, opened 2025-07-28
ui/widgets: use std::nullptr_t instead of nullptr_t

mmdbalkhi:
This fixes build issues on stricter environments (e.g. with musl libc), where `nullptr_t` is not available in the global namespace.


WhyNotHugo:
Thanks!

PR #266, opened 2025-10-07
Fix build errors related to std::nullptr_t

WhyNotHugo:
When building on Alpine Linux, a build error is produced due to nullptr_t being undefined. Importing <cstddef> fixes this.

It is unclear why the build error is not reproducible elsewhere; it may be due to different build flags or subtle differences in dependencies.

Regardless, usage of nullptr_t requires importing <cstddef>, so this change is correct everywhere, even if it is a no-op on some build configurations.


WhyNotHugo:
Actually, this is identical to https://github.com/desktop-app/lib_ui/pull/262

PR #296, opened 2026-04-11
Disable custom shadow painting for SeparatePanel on Wayland.

gearonixx:
**Problem**

On Wayland, SeparatePanel renders its own client-side shadow via `Shadow::paint()` and applies `st::callShadow.extend` as padding. However, on Wayland the compositor already provides window shadows. 

**Root Cause**

`createBorderImage()` and `initGeometry()` unconditionally applied `st::callShadow.extend` as shadow padding and drew the shadow, without checking whether the platform already provides compositor-level shadows.

how_it_was

**Fix**

fixed_state

**Affected file**

- `ui/widgets/separate_panel.cpp` (in `desktop-app/lib_ui`)

**Testing**

On Wayland SeparatePanel should display with compositor-provided shadow only

### DIFF
--- a/ui/widgets/separate_panel.cpp
+++ b/ui/widgets/separate_panel.cpp
@@ -1017,7 +1017,10 @@ void SeparatePanel::validateBorderImage() {
 }
 
 QPixmap SeparatePanel::createBorderImage(QColor color) const {
-	const auto shadowPadding = st::callShadow.extend;
+	const auto drawShadow = !::Platform::IsWayland();
+	const auto shadowPadding = drawShadow
+		? st::callShadow.extend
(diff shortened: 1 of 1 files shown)


gearonixx:
@ilya-fedin

ilya-fedin:
> However, on Wayland the compositor already provides window shadows.

I can't confirm that. With your patch the window becomes shadow-less on Wayland.

gearonixx:
You're right, thanks. Closing this.

The shadow I was seeing came from Hyprland compositor, so disabling the client-side shadow only looked correct locally and goes shadow-less elsewhere.

PR #297, opened 2026-04-16
Fix random heap as pixels noise

kukuruzka165:
image

(corrupted video)
image

(https://t.me/bg/kO4jyq55SFABAAAA0WEpcLfahXk?intensity=55&bg_color=bbbee8-c5c0e3&rotation=223 link in chat links preview)

### DIFF
--- a/ui/image/image_prepare.cpp
+++ b/ui/image/image_prepare.cpp
@@ -462,7 +462,15 @@ std::array<QImage, 4> PrepareCorners(
 	result.format = reader.format().toLower();
 	result.animated = reader.supportsAnimation()
 		&& (reader.imageCount() > 1);
+	result.image = QImage(size, QImage::Format_ARGB32_Premultiplied);
+	if (result.image.isNull()) {
+		return {};
+	}
+	result.image.fill(Qt::transparent);
 	if (!reader.read(&result.image) || result.image.isNull()) {
+		return {};
+	}
+	if (reader.error() != QImageReader::UnknownError) {
 		return {};
 	}
 	return result;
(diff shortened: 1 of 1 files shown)


ui/image/image_prepare.cpp
```diff
@@ -462,9 +462,17 @@ std::array<QImage, 4> PrepareCorners(
...
 		&& (reader.imageCount() > 1);
+	result.image = QImage(size, QImage::Format_ARGB32_Premultiplied);
+	if (result.image.isNull()) {
+		return {};
+	}
+	result.image.fill(Qt::transparent);
 	if (!reader.read(&result.image) || result.image.isNull()) {
 		return {};
 	}
+	if (reader.error() != QImageReader::UnknownError) {
```
ilya-fedin, line 473: This has no sense, if there's any error, QImageReader::read checked in the previous condition will return false

ui/image/image_prepare.cpp
```diff
@@ -462,9 +462,17 @@ std::array<QImage, 4> PrepareCorners(
 	result.format = reader.format().toLower();
 	result.animated = reader.supportsAnimation()
 		&& (reader.imageCount() > 1);
+	result.image = QImage(size, QImage::Format_ARGB32_Premultiplied);
```
ilya-fedin, line 465: Which format we're fixing for? If it's JPEG, I checked QJpegHandler code and it always re-allocates the image if either size or format isn't expected. And it never expects `QImage::Format_ARGB32_Premultiplied`. So if we're fixing for jpeg, this has no sense.

john-preston:
I suspect this should be fixed in 6.7.8, at least one place where we used non-filled initially image was fixed.

PR #304, opened 2026-06-07
feat(accessibility): support PageTab role and selected state on buttons

rezabakhshilaktasaraei:
## Summary
Exposes a custom tab control (Telegram's chat-folders strip) to screen readers by implementing the Qt 6 selection/attributes accessibility interfaces on the accessible container, plus the supporting `PageTab`/`PageTabList` state. Native on Qt 6; on Qt 5.15 it pairs with the qtbase backports in desktop-app/patches#254. Wired up in telegramdesktop/tdesktop#30772.

## What it does

**`Ui::Accessible::Widget`** (the `QAccessibleWidget` wrapper)
- Implements `QAccessibleSelectionInterface` for role `PageTabList`: the **current tab** is the selected item, answered from each child's `state().selected` — persistent and independent of keyboard focus.
- Implements `QAccessibleAttributesInterface`, exposed (via `interface_cast`) only when the widget reports an `accessibilityOrientation()` — returns the `Orientation` attribute so a vertical tab list is announced as vertical.
- Forwards accessible focus from the container to the active tab (`focusChild`).
- Drives `childCount`/`child`/`indexOfChild` from a new `accessibilityChildWidgets()` hook, so UIA sibling navigation follows the **visual** child order.

**`AbstractButton` / `SideBarButton`**
- `setIsPageTab()` / `isPageTab()`; `accessibilityRole()` returns `PageTab`.
- `accessibilityState()` **merges** the base state (keeps `pressed`) and, for a page tab, reports `selectable` + `selected = _active`.
- `setActive()` announces selection with a dedicated **`SelectionAdd`/`SelectionRemove`** event.
- `setDisabled()` keeps `QWidget::setEnabled()` in sync, so a disabled button is reported as disabled from the real widget state (`QAccessibleWidget`) rather than a custom field — this fixes disabled announcement for every button.

**`VerticalLayout`**
- Overrides `accessibilityChildWidgets()` (visual order) and emits `QAccessible::ObjectReorder` from `reorderRows()` — drag-reorder keeps the accessible sibling order in sync.

**`RpWidget` / `AccessibilityState`**
- New reusable hooks: `accessibilityChildWidgets()` and `accessibilityOrientation()` (default unset).
- New `selectable` state bit.

## Why selection is decoupled from focus
Qt's Windows bridge derives both `UIA_HasKeyboardFocus` and (in its default `PageTab` path) `IsSelected` from the same `state().focused` bit. Answering selection from a real `QAccessibleSelectionInterface` (reading `state().selected`) keeps the active tab `SELECTED` persistently while `HasKeyboardFocus` stays honest. On Qt 5.15 the bridge consults the interfaces via desktop-app/patches#254; on Qt 6 it's native.

### DIFF
--- a/ui/rp_widget.cpp
+++ b/ui/rp_widget.cpp
@@ -332,6 +332,7 @@ void AccessibilityState::writeTo(QAccessible::State &state) {
 	state.checked = checked ? 1 : 0;
 	state.pressed = pressed ? 1 : 0;
 	state.readOnly = readOnly ? 1 : 0;
+	state.selectable = selectable ? 1 : 0;
 	state.selected = selected ? 1 : 0;
 }
 
@@ -501,6 +502,10 @@ void RpWidget::accessibilityDoAction(const QString &name) {
 
 int RpWidget::accessibilityChildCount() const {
 	return -1;
+}
+
+std::vector<not_null<QWidget*>> RpWidget::accessibilityChildWidgets() const {
+	return {};
 }
 
 RpWidget *RpWidget::accessibilityParent() const {
--- a/ui/rp_widget.cpp
+++ b/ui/rp_widget.cpp
@@ -508,6 +508,10 @@ std::vector<not_null<QWidget*>> RpWidget::accessibilityChildWidgets() const {
 	return {};
 }
 
+std::optional<Qt::Orientation> RpWidget::accessibilityOrientation() const {
+	return std::nullopt;
+}
+
 RpWidget *RpWidget::accessibilityParent() const {
 	return nullptr;
 }
--- a/ui/rp_widget.h
+++ b/ui/rp_widget.h
@@ -422,6 +422,11 @@ public:
 	// use the default QWidget enumeration.
 	[[nodiscard]] virtual std::vector<not_null<QWidget*>> accessibilityChildWidgets() const;
 
+	// Orientation of a tab-control container (PageTabList), exposed to UIA so a
+	// screen reader can announce a horizontal/vertical tab list. nullopt (the
+	// default) means the widget reports no orientation.
+	[[nodiscard]] virtual std::optional<Qt::Orientation> accessibilityOrientation() const;
+
 	[[nodiscard]] virtual RpWidget *accessibilityParent() const;
 	[[nodiscard]] virtual QAccessibleInterface* accessibilityChildInterface(int index) const;
 	[[nodiscard]] virtual QString accessibilityChildName(int index) const;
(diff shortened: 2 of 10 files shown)


ui/rp_widget.cpp
```diff
@@ -333,6 +333,12 @@ void AccessibilityState::writeTo(QAccessible::State &state) {
 	state.pressed = pressed ? 1 : 0;
 	state.readOnly = readOnly ? 1 : 0;
 	state.selected = selected ? 1 : 0;
+	// Only ever set focused, never clear it, so the real keyboard-focus
+	// state from QAccessibleWidget is preserved. Used to expose the active
+	// tab as focused, the way QTabBar marks its current tab.
```
ilya-fedin: QAccessibleTabBar seem to manipulate with focused state only because the child items aren't real widgets. Since you're changing RpWidget, I assume they're real widgets in our case and you can use setFocus instead.

rezabakhshilaktasaraei: Thanks — the final approach addresses this more directly than `setFocus` would: it no longer manipulates `focused` at all.

The active tab now reports `state.selected` persistently — the same field Qt's own `QAccessibleTabButton` sets for the current tab (`s.selected = s.focused`). Real keyboard focus is untouched: `HasKeyboardFocus` follows actual focus, and the app already uses real `setFocus()` for keyboard navigation of the strip (`FiltersMenu::applyFilterAt`).

I didn't go `setFocus()`-only because the current tab has to read as *selected* even when the strip doesn't hold keyboard focus (e.g. reviewing it from the chat list), and `setFocus()` only reports focus while the widget literally has it. Driving it from `selected` gives the persistent selection while keeping focus honest — and avoids a side effect the focused-based version had: Qt's UIA bridge derives **both** `HasKeyboardFocus` and a PageTab's `IsSelected` from `state().focused`, so faking `focused` made the active tab report keyboard focus persistently. The qtbase side (desktop-app/patches#254) reads `IsSelected` from `state().selected` to match.

ui/rp_widget.h
```diff
@@ -374,7 +374,9 @@ struct AccessibilityState {
 	bool checked : 1 = false;
 	bool pressed : 1 = false;
 	bool readOnly : 1 = false;
+	bool selectable : 1 = false;
 	bool selected : 1 = false;
+	bool disabled : 1 = false;
```
ilya-fedin: You're breaking the "Don't override what QAccessibleWidget handles" rule again. You must use QWidget::setEnabled instead.

rezabakhshilaktasaraei: Done — removed the faked `disabled` field. `AbstractButton::setDisabled()` now keeps `QWidget::setEnabled()` in sync, so the disabled state comes from the real widget and `QAccessibleWidget` reports it (no override). Bonus: every disabled button now announces correctly — verified with NVDA, it reads as "unavailable".

PR #307, opened 2026-06-17
Fix stuck emoji after inserting it via Windows emoji picker.

1nv:
See https://github.com/telegramdesktop/tdesktop/issues/29806


john-preston:
Thanks!

PR #320, opened 2026-07-15
Mac overscroll

teidesu:
> **full disclosure**: i wrote exactly zero lines of code myself, but i reviewed and tested the changes thoroughly (i think (maybe))

ok so overscroll behavior currently implemented in ElasticScroll feels very bad and i asked fable to port the chromium(blink) macos overscroll implementation over here. 

that alone already improved things a lot, but then we also reworked the pull-action stuff (pull-to-next-channel in history and pull-to-stories in chat list), making them feel a lot nicer and more responsive - instead of the log curve it was using.

currently all changes are gated behind `Q_OS_MACOS` (because i dont have any win/linux device to test them), but most of them *should* be portable and work on other platforms with little to no changes. (however kinetic scrolling on both linux and windows needs extra care i believe)

**note:** the changes in the `lib_ui` are just the "core" and require some slight changes in the actual consumers of the main app - see https://github.com/telegramdesktop/tdesktop/pull/30990

before/after (videos dont really do justice, it should be compared side-by-side with an actual trackpad, but i tried my best):

https://github.com/user-attachments/assets/c89f03c3-ea60-4555-898b-444e9e4de180

https://github.com/user-attachments/assets/ec27cfe3-ebb7-4355-801d-e9339e15f0fa

https://github.com/user-attachments/assets/bff25e46-fced-494f-b08a-861484528a30

https://github.com/user-attachments/assets/c1c536ce-3720-4f9f-aea0-f6aeb5469c44

https://github.com/user-attachments/assets/e8e7a18f-f5a2-4447-9730-4883316b5826

https://github.com/user-attachments/assets/daa44912-566b-456c-a3b2-993e4574b78b

### DIFF
--- a/ui/widgets/elastic_scroll.cpp
+++ b/ui/widgets/elastic_scroll.cpp
@@ -66,6 +66,25 @@ constexpr auto kTouchOverscrollMultiplier = 2;
 constexpr auto kLogA = 16.;
 constexpr auto kLogB = 10.;
 
+#ifdef Q_OS_MAC
+// Rubber-band physics matching WebKit / Chromium
+// (blink::ElasticOverscrollControllerExponential), which in turn
+// replicate the native macOS feel: linear stretch = accumulated
+// overscroll / stiffness while dragging, and a critically damped
+// spring x(t) = (x0 + v0 * amplitude * t) * e^(-stiffness * t / period)
+// on release, seeded with the tracked scroll velocity.
+constexpr auto kRubberBandStiffness = 20.;
+constexpr auto kRubberBandAmplitude = 0.31;
+constexpr auto kRubberBandPeriod = 1.6;
+constexpr auto kRubberBandMaxReturnDuration = crl::time(1000);
+constexpr auto kVelocityZeroingTimeout = crl::time(100);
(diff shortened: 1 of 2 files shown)


ui/widgets/elastic_scroll.cpp
```diff
@@ -514,6 +570,137 @@ void ElasticScroll::overscrollReturn() {
...
+}
+
+float64 ElasticScroll::overscrollPullDistance(int side) const {
+	if (!overscrollPullSide(side)) {
+		return 0.;
+	}
+	const auto distance = (side < 0)
+		? _overscrollPullFrom
+		: _overscrollPullTill;
+	return distance ? float64(distance) : 1e9;
```
23rd, line 602: `1e9` is good enough to be placed in `constexpr` as well.

23rd:
The comments are too wordy and verbose.

gearonixx:
I can reproduce the same overshoot on Linux x86_64 with the `QScroller` option enabled.

gearonixx:
PR: https://github.com/telegramdesktop/tdesktop/pull/30991

john-preston:
Thanks

PR #322, opened 2026-07-17
Fix stale link activation from checkboxes

### LINKED ISSUE #30982: Clicking the notification checkbox in the pin dialog also opens the underlying link
### Steps to reproduce

1. Open a group or channel where you have permission to pin messages.
2. Find or send a message with large preview containing a clickable external link, such as a YouTube link. In my case last time it was inline @vid bot youtube results for Lucky Official Trailer: https://www.youtube.com/watch?v=GAbT5qCTXR8
4. Right-click the message and select Pin.
5. In the pin confirmation dialog, click the notification checkbox text to disable notifications.
6. Make sure the checkbox text is displayed over the link in the message behind the dialog. Resize window if it helps to position it just right.

### Expected behaviour

The click should be handled exclusively by the pin confirmation dialog. It should only toggle the notification checkbox, and no element in the chat behind the dialog should receive the click.

### Actual behaviour

The notification checkbox is unchecked, but the same mouse click also passes through the dialog to the message behind it. If a link (or its whole clickable preview area) is located underneath the checkbox text, Telegram opens that link in the browser.

This is consistently reproducible when the notification checkbox text overlaps a clickable link in the underlying message.

GPT 5.6 guess: _the checkbox calls the global ClickHandler::pressed() without refreshing or clearing the active handler under the current cursor position._

### Operating system

Windows 11

### Version of Telegram Desktop

6.9.3 x64

### Installation source

Static binary from official website

### Crash ID

_No response_

### Logs

```shell

```

gazakov:
Fixes telegramdesktop/tdesktop#30982.

When a checkbox dialog opens over a link without mouse movement, the old link can remain the active ClickHandler. The checkbox then presses and activates that stale handler on release.

This refreshes the handler from the current checkbox position before recording the press.

Tested by building the lib_ui target in the TDesktop 7.0.1 macOS arm64 Release configuration.

### DIFF
--- a/ui/widgets/checkbox.cpp
+++ b/ui/widgets/checkbox.cpp
@@ -800,6 +800,11 @@ void Checkbox::paintEvent(QPaintEvent *e) {
 
 void Checkbox::mousePressEvent(QMouseEvent *e) {
 	RippleButton::mousePressEvent(e);
+	const auto state = getTextState(e->pos());
+	if (state.link != ClickHandler::getActive()) {
+		ClickHandler::setActive(state.link, this);
+		update();
+	}
 	ClickHandler::pressed();
 }
 


john-preston:
Thanks!

gazakov:
Thanks! I see this was fixed with a cleaner reuse of mouseMoveEvent. Closing as superseded.

PR #324, opened 2026-07-20
smoother scrolling on macos

teidesu:
this is part of an effort to make scrolling smoother on macos (together with telegramdesktop/tdesktop#31023), and it kinda worked out.

- **use exact native scroll deltas on macos** — qt rounds NSEvent's fractional `scrollingDelta` to whole logical pixels when filling `pixelDelta()`. install a local NSEvent monitor that records the exact fractional deltas and match them back to each QWheelEvent by timestamp (validated against the rounded delta so an unrelated event can't be mistaken for it).
- **carry sub-pixel scroll remainder across wheel events** — scroll positions are whole pixels, so the fractional part of each delta used to be dropped at every event. keep the remainder across a gesture instead, and expose the sub-pixel offset (`subpixelScrollValue()`) so content can render translated by the fraction. on macos the offset is interpolated over ~one frame and snapped to the device-pixel grid. reset on programmatic scrolls and gesture begin.

note: the code in this pr was written partially by claude fable 5 and partially by gpt-5.6-sol.


ui/widgets/elastic_scroll.cpp
```diff
@@ -1027,27 +1029,65 @@ bool ElasticScroll::handleWheelEvent(not_null<QWheelEvent*> e, bool touch) {
...
+			: (total - delta);
+#ifdef Q_OS_MAC
+		const auto interpolate = !touch
+			&& (phase != Qt::NoScrollPhase)
+			&& !_overscroll
+			&& (preciseTarget >= 0.)
+			&& (preciseTarget <= maximum);
+#else // Q_OS_MAC
+		const auto interpolate = false;
+#endif // Q_OS_MAC
```
ilya-fedin:
```suggestion
		const auto interpolate = Platform::IsMac()
			&& !touch
			&& (phase != Qt::NoScrollPhase)
			&& !_overscroll
			&& (preciseTarget >= 0.)
			&& (preciseTarget <= maximum);
```

teidesu:
update: pushed a second approach as a new commit (14efbbd)

- **commits 1–2** (original): a local NSEvent monitor records exact fractional `scrollingDelta`s and matches them back to each QWheelEvent by timestamp. self-contained, but heuristic (match can fail) and macos-only.
- **commit 14efbbd** (new): drops the monitor entirely and reads `pixelDeltaF()` — a strictly additive getter from a qtbase patch (desktop-app/patches#262) that carries the exact fractional delta through the QPA. consumed via `if constexpr (requires { e->pixelDeltaF(); })`, so this still compiles and gracefully degrades to rounded deltas against stock qt.

imo the second approach is cleaner — no global event monitor, no timestamp-matching heuristic, and the same plumbing would work for wayland's fractional deltas too.

teidesu:
ok after further testing it turned out that:
- `NSEvent`-s don't actually carry fractional deltas at all, the fractional deltas i saw were from the default 110% scaling (bruh)
- the current sub-pixel interpolation actually *increases* the perceived latency while actually dragging
- blind testing sub-pixel scrolling (with interpolation only applied to kinetic scrolling, see prev. point) doesn't seem to make any meaningful perceived difference (ime, at least)

the primary fix was likely the displaylink rhi fix, these are likely placebo and instead i believe the microstutters themselves (heavy work on main thread) should be tackled

PR #327, opened 2026-07-21
Replace QScroller with custom KineticScroller.

poli0iq:
`QScroller` has:
1. A bad inertia curve (all of the default ones).
2. A bad initial acceleration estimation.

These can't be properly tuned, so the only option is to reimplement `QScroller`.

This implementation's maths is based on Chromium and GTK4.

This also replaces the screen refresh rate heuristics with proper frame scheduling (vsync-accurate).

`KineticScroller` also exposes velocity directly, allowing more precise overscroll behavior without relying on estimated velocity from rounded per-tick deltas (fixing the spring suppression in the slow fling tails).

### DIFF
--- a/ui/widgets/elastic_scroll.cpp
+++ b/ui/widgets/elastic_scroll.cpp
@@ -8,13 +8,11 @@
 
 #include "ui/painter.h"
 #include "ui/ui_utility.h"
-#include "base/options.h"
 #include "base/qt/qt_common_adapters.h"
 #include "styles/style_widgets.h"
 
 #include <QtGui/QWindow>
 #include <QtCore/QtMath>
-#include <QtWidgets/QScroller>
 #include <QtWidgets/QApplication>
 
 namespace Ui {
@@ -425,36 +423,6 @@ ElasticScroll::ElasticScroll(
 	) | rpl::on_next([=](int from) {
 		tryScrollTo(from, false);
 	}, _bar->lifetime());
-
-	static const auto &OptionQScroller = base::options::lookup<bool>(
-		kOptionQScroller);
-
-	rpl::single(
-		rpl::empty
-	) | rpl::then(
-		OptionQScroller.changes()
-	) | rpl::on_next([=] {
-		if (OptionQScroller.value()) {
-			_scroller = QScroller::scroller(this);
-			SetupScrollerPhysics(_scroller);
-		} else if (_scroller) {
-			QObject deleter;
-			_scroller->setParent(&deleter);
-		}
-
-		if (!_scroller) {
-			return;
-		}
-
-		connect(
-			_scroller,
-			&QScroller::stateChanged,
-			[=](QScroller::State state) {
-				if (state == QScroller::Scrolling) {
-					ScrollerStopper::Instance().activate(_scroller);
-				}
-			});
-	}, lifetime());
 }
 
 ElasticScroll::~ElasticScroll() {
@@ -634,15 +602,6 @@ void ElasticScroll::overscrollSpringStart(int side) {
 		0.,
 		1.,
 		kRubberBandMaxReturnDuration);
-}
-
-void ElasticScroll::overscrollBounce(int side, float64 velocity) {
-	_overscrollReturning = true;
-	_ignoreMomentumFromOverscroll = side;
-	_movement = Movement::Returning;
-	_wheelVelocity = velocity;
-	_wheelVelocityTime = crl::now();
-	overscrollSpringStart(side);
 }
 
 void ElasticScroll::overscrollSpringUpdate() {
@@ -864,85 +823,6 @@ bool ElasticScroll::eventHook(QEvent *e) {
 		}
 		return true;
 	}
-	switch (e->type()) {
-	case QEvent::ScrollPrepare: {
-		QScrollPrepareEvent *se = static_cast<QScrollPrepareEvent *>(e);
-		se->setViewportSize(QSizeF(size()));
-		se->setContentPosRange(QRectF(
-			0,
-			0,
-			_vertical ? 0 : _state.fullSize - width(),
-			_vertical ? _state.fullSize - height() : 0));
-		se->setContentPos(QPointF(
-			_vertical ? 0 : _state.visibleFrom,
-			_vertical ? _state.visibleFrom : 0));
-		se->accept();
-		return true;
-	}
-	case QEvent::Scroll: {
-		QScrollEvent *se = static_cast<QScrollEvent *>(e);
-		const auto state = _scroller->state();
-		const auto phase = state == QScroller::Pressed
-			? Qt::ScrollBegin
-			: state == QScroller::Dragging
-			? Qt::ScrollUpdate
-			: state == QScroller::Scrolling
-			? Qt::ScrollMomentum
-			: Qt::ScrollEnd;
-		const auto pixels
-			= (se->contentPos() + se->overshootDistance()).toPoint();
-		const auto delta
-			= -(_state.visibleFrom - (_vertical ? pixels.y() : pixels.x()));
-		// Capture the fling velocity before this event is tracked: at the
-		// boundary QScroller clamps the delta, so the sample it produces
-		// underestimates the speed the fling actually hit the edge with.
-		const auto velocity = (_wheelVelocityTime
-			&& (crl::now() - _wheelVelocityTime <= kVelocityZeroingTimeout))
-			? _wheelVelocity
-			: 0.;
-		const auto weak = base::make_weak(this);
-		const auto result = handleScrollEvent(phase, delta);
-		if (!weak) {
-			return true;
-		}
-		if (phase == Qt::ScrollMomentum
-			&& velocity != 0.
-			&& _scroller
-			&& _scroller->state() == QScroller::Scrolling
-			&& !_overscrollReturnAnimation.animating()
-			&& _overscroll == currentOverscrollDefault()) {
-			// QScroller's overshoot is disabled, so a fling just stops
-			// dead at the boundary of the range QScroller knows about.
-			const auto side = (velocity < 0.) ? -1 : 1;
-			const auto target = _state.visibleFrom + side;
-			if (willScrollTo(target) != target) {
-				if (side > 0 && requestBottomContent(1)) {
-					// The boundary wasn't a real edge: more content was
-					// appended below, let the fling continue into it.
-					if (weak) {
-						ResendScrollerPrepare(_scroller);
-					}
-					return true;
-				}
-				if (!weak) {
-					return true;
-				}
-				const auto &allowed = (side < 0)
-					? _overscrollAllowFrom
-					: _overscrollAllowTill;
-				if (overscrollSpringSide(side)
-					&& (!allowed || allowed())) {
-					// A real edge: hand the residual velocity over to the
-					// rubber-band spring for the bounce.
-					_scroller->stop();
-					_wheelPos = {};
-					overscrollBounce(side, velocity);
-				}
-			}
-		}
-		return result;
-	}
-	}
 	return RpWidget::eventHook(e);
 }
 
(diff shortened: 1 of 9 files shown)


ui/widgets/scroll_area.cpp
```diff
@@ -564,14 +498,12 @@ ScrollArea::ScrollArea(
...
 	) | rpl::on_next([=] {
-		if (OptionQScroller.value()) {
-			_scroller = QScroller::scroller(this);
-			SetupScrollerPhysics(_scroller);
+		if (OptionKineticScroller.value()) {
+			_scroller = new KineticScroller(this);
 		} else if (_scroller) {
-			QObject deleter;
-			_scroller->setParent(&deleter);
+			delete _scroller;
```
ilya-fedin: If we have access to constructor and destructor then use a smart pointer
```suggestion
		_scroller = OptionKineticScroller.value()
			? std::make_unique<KineticScroller>(this);
			: nullptr;
```

ui/widgets/scroll_area.cpp
```diff
@@ -795,30 +727,40 @@ bool ScrollArea::viewportEvent(QEvent *e) {
...
-						? QPoint(
-							unmultiplied.x() * std::max(width(), 120) / 120.,
-							unmultiplied.y() * std::max(height(), 120) / 120.)
-						: unmultiplied;
+					_scroller->handleInput(
+						KineticScroller::InputPress,
+						_wheelPos,
+						crl::now());
+				}
+				// The press event's own delta counts too: a short fling
```
ilya-fedin: So we're resurrecting ScrollBegin's delta here while ignoring it multiple lines above if we're in KineticScroller::Scrolling state?

ui/widgets/elastic_scroll.cpp
```diff
@@ -426,20 +425,18 @@ ElasticScroll::ElasticScroll(
...
-		if (OptionQScroller.value()) {
-			_scroller = QScroller::scroller(this);
-			SetupScrollerPhysics(_scroller);
+		if (OptionKineticScroller.value()) {
+			_scroller = new KineticScroller(this);
 		} else if (_scroller) {
-			QObject deleter;
-			_scroller->setParent(&deleter);
+			delete _scroller;
 		}
```
ilya-fedin: Smart pointer too.
```suggestion
		_scroller = OptionKineticScroller.value()
			? std::make_unique<KineticScroller>(this);
			: nullptr;
```

ui/widgets/scroll_area.cpp
poli0iq: These are different deltas. The one ignored above is the leaked macOS momentum stream (qnsview_mouse.mm can emit `ScrollBegin` straight from `momentumPhase == Began` when it misses the handoff), which we're already simulating ourselves.

On Wayland, though, `ScrollBegin` always has a zero delta, so it's the first `ScrollUpdate` that triggers the press, and its delta is a velocity sample we shouldn't drop.

ilya-fedin: How it happens ScrollUpdate triggers the press? ScrollBegin should trigger InputPress and the next ScrollUpdate triggers InputMove.

poli0iq: `_wheelPos` being null means no gesture is open. Normally `ScrollBegin` opens it with `InputPress`, then each `ScrollUpdate` sends `InputMove`. But the first event we see can be a `ScrollUpdate`, for example when the check above swallowed a real begin during a fling. In that case `wasNull` is true on the update, so the press comes from it. And unlike a begin, the update carries a real scroll delta, which we then send as the first `InputMove`: a short fling has only a few velocity samples, dropping one would noticeably weaken it.

ilya-fedin: > But the first event we see can be a `ScrollUpdate`, for example when the check above swallowed a real begin during a fling.

You just said ScrollBegin always has zero delta on Wayland? If that's right, then this change has zero effect and should be reverted. If not, then remove the macOS workaround as macOS users aren't the expected users of this feature anyway.

poli0iq: Alright, I've debugged raw events and this path is indeed useless, the problem was related to Bluetooth buffering some events (leading to losing some short flings because we use `crl::now()` timestamps instead of `QWheelEvent::timestamp()`).

(yeah I use a bluetooth touchpad)

ui/widgets/scroll_area.cpp
```diff
@@ -564,26 +498,21 @@ ScrollArea::ScrollArea(
...
+		OptionKineticScroller.changes()
 	) | rpl::on_next([=] {
-		if (OptionQScroller.value()) {
-			_scroller = QScroller::scroller(this);
-			SetupScrollerPhysics(_scroller);
-		} else if (_scroller) {
-			QObject deleter;
-			_scroller->setParent(&deleter);
-		}
-
```
ilya-fedin, line 605: could the empty line be returned?

ui/widgets/elastic_scroll.cpp
```diff
@@ -426,32 +425,27 @@ ElasticScroll::ElasticScroll(
...
+		OptionKineticScroller.changes()
 	) | rpl::on_next([=] {
-		if (OptionQScroller.value()) {
-			_scroller = QScroller::scroller(this);
-			SetupScrollerPhysics(_scroller);
-		} else if (_scroller) {
-			QObject deleter;
-			_scroller->setParent(&deleter);
-		}
-
```
ilya-fedin, line 444: and here

ui/widgets/scroll_area.cpp
ilya-fedin: crl::now() was used only because QFlickGestureRecognizer uses QElapsedTimer so I mirrored it. If event timestamps are in fact better then we can use them.

ilya-fedin:
Last stylistic thing I see right now: both new .cpp and .h seem to use spaces rather than tabs.

And to technical things:
1. An annoying thing with QScroller is that it works in absolute coordinates, we have to convert deltas to absolute positions and back, just to then bug on loading more messages. Given that we change it, couldn't we get rid of this too and just directly operate on delta? Or is it crucial to the logic to know scroll area size and position?
2. If possible, I believe it would be nice to get rid of the handleWheelEvent/handleScrollEvent separation just like it was before QScroller introduction. We had to use scroll events with it, but we're not forced anymore with an in-tree API I believe.
3. Is there any sense to have separate ScrollerStopper now or would it be more effectively done by KineticScroller itself?

poli0iq:
> Last stylistic thing I see right now: both new .cpp and .h seem to use spaces rather than tabs.

I don't think they do.

> And to technical things:
> 
> 1. An annoying thing with QScroller is that it works in absolute coordinates, we have to convert deltas to absolute positions and back, just to then bug on loading more messages. Given that we change it, couldn't we get rid of this too and just directly operate on delta? Or is it crucial to the logic to know scroll area size and position?
> 
> 2. If possible, I believe it would be nice to get rid of the handleWheelEvent/handleScrollEvent separation just like it was before QScroller introduction. We had to use scroll events with it, but we're not forced anymore with an in-tree API I believe.
> 
>  3. Is there any sense to have separate ScrollerStopper now or would it be more effectively done by KineticScroller itself?

All three implemented (as separate commits for now, lmk if I should fixup).

ui/widgets/scroll_area.cpp
```diff
@@ -656,62 +656,50 @@ bool ScrollArea::viewportEvent(QEvent *e) {
...
+			const auto delta = multiply
+				? QPoint(
+					unmultiplied.x() * std::max(width(), 120) / 120.,
+					unmultiplied.y() * std::max(height(), 120) / 120.)
+				: unmultiplied;
+			const auto horizontal = horizontalScrollBar();
+			const auto vertical = verticalScrollBar();
+			const auto wasX = horizontal->value();
+			const auto wasY = vertical->value();
+			horizontal->setValue(wasX - delta.x());
```
ilya-fedin: Why this is done? Now that KineticScroller produces wheel events, QScrollArea should handle it just fine by itself!

ui/widgets/kinetic_scroller.cpp
```diff
@@ -131,46 +126,58 @@ void KineticScroller::flick(crl::time timestamp) {
...
 	}
 }
 
+void KineticScroller::sendWheel(Qt::ScrollPhase phase, QPoint pixel) {
+	const auto global = QCursor::pos();
+	auto e = QWheelEvent(
+		_target->mapFromGlobal(global),
+		global,
+		pixel,
+		pixel * kPixelToAngleDelta,
```
ilya-fedin: Maybe count kinetic effect from both pixel and angle delta instead of emulating angle delta from pixel delta? Or is it not possible?

ui/widgets/scroll_area.h
```diff
@@ -9,7 +9,7 @@
 #include "ui/rp_widget.h"
 #include "ui/ui_utility.h"
 #include "ui/effects/animations.h"
-#include "ui/widgets/kinetic_scroller.h"
+#include "ui/widgets/kinetic_scroller.h" // For kOptionKineticScroller.
```
ilya-fedin: It's not used here?

ui/widgets/scroll_area.cpp
```diff
@@ -656,41 +635,38 @@ bool ScrollArea::viewportEvent(QEvent *e) {
...
-			// unchanged" is not enough: sub-quantum ticks legitimately
-			// leave it unchanged mid-range while the fraction accumulates.
-			const auto delta = (horizontal ? angle.x() : angle.y())
-				* (ev->inverted() ? -1 : 1);
-			if (bar->value() == was
-				&& ((delta > 0 && was == bar->minimum())
-					|| (delta < 0 && was == bar->maximum()))) {
-				// The fling ran into the end of the range.
-				_scroller->stop();
+		if (ev->phase() == Qt::ScrollMomentum) {
```
ilya-fedin: Why do we need to special case kinetic scroller here?

ui/widgets/kinetic_scroller.cpp
```diff
@@ -96,13 +195,46 @@ void KineticScroller::stop() {
...
+		// Retargeting must not leave the old target's momentum stream
+		// dangling: flush its final ScrollEnd first. May destroy `this`.
+		_emitted = false;
+		const auto weak = QPointer<KineticScroller>(this);
+		sendWheel(Qt::ScrollEnd, {}, {});
+		if (!weak) {
+			return false;
+		}
+	}
+	_target = target.get();
```
ilya-fedin, line 223: Why do we care about widgets?

ui/widgets/elastic_scroll.cpp
```diff
@@ -887,6 +874,9 @@ bool ElasticScroll::handleWheelEvent(not_null<QWheelEvent*> e, bool touch) {
 		return true;
 	}
 	const auto phase = e->phase();
+	const auto scroller = (phase != Qt::NoScrollPhase)
```
ilya-fedin: Why do we need to special case kinetic scroller here?

ui/widgets/kinetic_scroller.cpp
poli0iq: Because synthetic events are not `spontaneous()` and therefore don't go through [this path](https://invent.kde.org/qt/qt/qtbase/-/blob/6.11.2/src/widgets/kernel/qapplication.cpp?ref_type=heads#L2811).

ilya-fedin: You can fix that with QWindowSystemInterface::handleWheelEvent like ui_windows_direct_manipulation.cpp does?

ui/widgets/elastic_scroll.cpp
```diff
@@ -958,36 +945,19 @@ bool ElasticScroll::handleWheelEvent(not_null<QWheelEvent*> e, bool touch) {
...
-			if (!weak) {
-				return true;
-			}
-		}
 	}
 	const auto momentum = (phase == Qt::ScrollMomentum)
 		|| (phase == Qt::ScrollEnd);
 	trackWheelVelocity(phase, delta, crl::time(e->timestamp()));
-	if (_scroller && _scroller->state() == KineticScroller::Scrolling) {
+	const auto scroller = KineticScroller::For(this);
```
ilya-fedin: Why do we need to special case kinetic scroller here?

ui/widgets/kinetic_scroller.cpp
```diff
@@ -0,0 +1,479 @@
...
+	crl::time _inputReceivedAt = 0;
+
+	Velocity _flickVelocity;
+	QPointF _flickEmittedPixel;
+	QPointF _flickEmittedAngle;
+	crl::time _flickStarted = 0;
+
+};
+
+QPointer<KineticScroller> GlobalScroller;
```
ilya-fedin: Move inside InstallKineticScroller as a static

ui/widgets/kinetic_scroller.cpp
```diff
@@ -0,0 +1,479 @@
...
+		const auto guardedWindow = QPointer<QWindow>(window);
+		if (e->phase() == Qt::NoScrollPhase) {
+			const auto generation = ++_generation;
+			interrupt(true);
+			return !guardedWindow || generation != _generation;
+		} else if (e->source() == Qt::MouseEventSynthesizedByApplication) {
+			return false;
+		}
+		const auto generation = ++_generation;
+		if (!OptionKineticScroller.value()) {
```
ilya-fedin: Better check that right inside InstallKineticScroller, you can then do like that:
```
static const auto lifetime = rpl::single(
	rpl::empty
) | rpl::then(
	OptionKineticScroller.changes()
) | rpl::on_next([] {
	static base::unique_qptr<KineticScroller> Scroller;
	Scroller = OptionKineticScroller.value()
		? base::make_unique_q<KineticScroller>(qApp)
		: nullptr;
	if (!Scroller) {
		return;
	}
	qApp->installEventFilter(Scroller.get());
});

john-preston:
lfg

PR #335, opened 2026-08-07
Name the tooltip close button for screen readers

rezabakhshilaktasaraei:
The close button in MakeTooltipWithClose had no accessible name, so screen readers announced it as just "button". Now it uses Integration::phraseButtonClose(), the same way the window title close button does.

### DIFF
--- a/ui/widgets/tooltip.cpp
+++ b/ui/widgets/tooltip.cpp
@@ -6,6 +6,7 @@
 //
 #include "ui/widgets/tooltip.h"
 
+#include "ui/integration.h"
 #include "ui/ui_utility.h"
 #include "ui/painter.h"
 #include "ui/platform/ui_platform_utility.h"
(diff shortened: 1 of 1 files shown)


ui/widgets/tooltip.cpp
```diff
@@ -527,6 +528,7 @@ object_ptr<RpWidget> MakeTooltipWithClose(
 	const auto button = CreateChild<IconButton>(
 		result.data(),
 		closeSt);
+	button->setAccessibleName(closeLabel);
```
ilya-fedin: `Integration::phraseButtonClose` doesn't work?

rezabakhshilaktasaraei: It does - the button just had no accessible name at all, so NVDA read it as bare "button". Switched to phraseButtonClose(), dropped the parameter and the tdesktop-side PR.

PR #339, opened 2026-08-14
Fix X11 touchpad scrolling in elastic views

catap:
Normalize smooth wheel input while preserving ordinary wheel behavior; retain fractional deltas and report the visible viewport to QScroller.

### DIFF
--- a/ui/ui_utility.cpp
+++ b/ui/ui_utility.cpp
@@ -7,6 +7,7 @@
 #include "ui/ui_utility.h"
 
 #include "base/platform/base_platform_info.h"
+#include "base/qt/qt_common_adapters.h"
 #include "ui/integration.h"
 #include "ui/style/style_core.h"
 
@@ -22,6 +23,7 @@ namespace {
 
 constexpr auto kDefaultWheelScrollLines = 3;
 constexpr auto kMagicScrollMultiplier = 2.5;
+constexpr auto kX11SmoothScrollMultiplier = 5.;
 
 class WidgetCreator : public QWidget {
 public:
@@ -272,18 +274,38 @@ QPointF ScrollDeltaF(not_null<QWheelEvent*> e, bool touch) {
 			style::ConvertScaleExact(point.x()),
 			style::ConvertScaleExact(point.y()));
 	};
+	const auto pixelDelta = e->pixelDelta();
+	const auto angleDelta = e->angleDelta();
 #if QT_VERSION >= QT_VERSION_CHECK(6, 2, 0)
 	using QInputDevice::Capability::PixelScroll;
-	if (touch || e->device()->capabilities().testFlag(PixelScroll)) {
+	const auto pixelInput = touch
+		|| e->device()->capabilities().testFlag(PixelScroll);
 #else // Qt >= 6.2.0
-	if (!e->pixelDelta().isNull()) {
+	const auto pixelInput = !pixelDelta.isNull();
 #endif // Qt < 6.2.0
-		return convert(e->pixelDelta())
+	auto x11Touchpad = false;
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+	x11Touchpad = e->device()
+		&& e->device()->type() == base::TouchDevice::TouchPad;
+#endif // Qt >= 6.0.0
+	const auto x11Smooth = !touch
+		&& ::Platform::IsX11()
+		&& (pixelInput || !pixelDelta.isNull() || x11Touchpad);
+	if (x11Smooth && !angleDelta.isNull()) {
+		return (convert(angleDelta)
+			* QApplication::wheelScrollLines()
+			* kX11SmoothScrollMultiplier)
+			/ float64(QWheelEvent::DefaultDeltasPerStep);
+	}
+	if (pixelInput || (x11Smooth && !pixelDelta.isNull())) {
+		return (::Platform::IsX11() && !touch
+			? QPointF(pixelDelta)
+			: convert(pixelDelta))
 			* ((::Platform::IsWayland() && !touch)
 				? kMagicScrollMultiplier
 				: 1.);
 	}
-	return (convert(e->angleDelta()) * QApplication::wheelScrollLines())
+	return (convert(angleDelta) * QApplication::wheelScrollLines())
 		/ float64(kPixelToAngleDelta * kDefaultWheelScrollLines);
 }
 
--- a/ui/widgets/elastic_scroll.cpp
+++ b/ui/widgets/elastic_scroll.cpp
@@ -867,7 +867,7 @@ bool ElasticScroll::eventHook(QEvent *e) {
 	switch (e->type()) {
 	case QEvent::ScrollPrepare: {
 		QScrollPrepareEvent *se = static_cast<QScrollPrepareEvent *>(e);
-		se->setViewportSize(QSizeF(viewport()->size()));
+		se->setViewportSize(QSizeF(size()));
 		se->setContentPosRange(QRectF(
 			0,
 			0,
@@ -1029,21 +1029,29 @@ bool ElasticScroll::handleWheelEvent(not_null<QWheelEvent*> e, bool touch) {
 	const auto guard = gsl::finally([&] {
 		_lastScroll = now;
 	});
-	auto unmultiplied = ScrollDelta(e, touch);
+	if (phase == Qt::ScrollBegin) {
+		_wheelDeltaRemainder = {};
+	}
+	auto exactPixels = ScrollDeltaF(e, touch);
 	if (ownAxisLocked) {
 		if (_vertical) {
-			unmultiplied.setX(0);
+			exactPixels.setX(0.);
+			_wheelDeltaRemainder.setX(0.);
 		} else {
-			unmultiplied.setY(0);
+			exactPixels.setY(0.);
+			_wheelDeltaRemainder.setY(0.);
 		}
 	}
 	const auto multiply = e->modifiers()
 		& (Qt::ControlModifier | Qt::ShiftModifier);
-	const auto pixels = multiply
-		? QPoint(
-			(unmultiplied.x() * std::max(width(), 120) / 120.),
-			(unmultiplied.y() * std::max(height(), 120) / 120.))
-		: unmultiplied;
+	if (multiply) {
+		exactPixels = QPointF(
+			exactPixels.x() * std::max(width(), 120) / 120.,
+			exactPixels.y() * std::max(height(), 120) / 120.);
+	}
+	exactPixels += _wheelDeltaRemainder;
+	const auto pixels = exactPixels.toPoint();
+	_wheelDeltaRemainder = exactPixels - QPointF(pixels);
 	auto ignore = false;
 	auto delta = _vertical ? -pixels.y() : -pixels.x();
 	if (!ownAxisLocked
@@ -1092,7 +1100,7 @@ bool ElasticScroll::handleWheelEvent(not_null<QWheelEvent*> e, bool touch) {
 		case Qt::ScrollBegin:
 		case Qt::ScrollUpdate: {
 			if (phase == Qt::ScrollBegin
-				&& !pixels.isNull()
+				&& !exactPixels.isNull()
 				&& _scroller->state() == QScroller::Scrolling) {
 				// On macOS, when Qt loses the race detecting that a
 				// momentum phase follows the finger lift, the OS momentum
--- a/ui/widgets/elastic_scroll.h
+++ b/ui/widgets/elastic_scroll.h
@@ -334,6 +334,7 @@ private:
 
 	QPointer<QScroller> _scroller;
 	QPoint _wheelPos;
+	QPointF _wheelDeltaRemainder;
 
 	base::Timer _touchTimer;
 	base::Timer _touchScrollTimer;


ui/ui_utility.cpp
```diff
@@ -22,6 +23,7 @@ namespace {
 
 constexpr auto kDefaultWheelScrollLines = 3;
 constexpr auto kMagicScrollMultiplier = 2.5;
+constexpr auto kX11SmoothScrollMultiplier = 5.;
```
ilya-fedin, line 26: Why exactly that multilplier?

ui/ui_utility.cpp
```diff
@@ -272,18 +274,38 @@ QPointF ScrollDeltaF(not_null<QWheelEvent*> e, bool touch) {
...
-	if (!e->pixelDelta().isNull()) {
+	const auto pixelInput = !pixelDelta.isNull();
 #endif // Qt < 6.2.0
-		return convert(e->pixelDelta())
+	auto x11Touchpad = false;
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+	x11Touchpad = e->device()
+		&& e->device()->type() == base::TouchDevice::TouchPad;
+#endif // Qt >= 6.0.0
+	const auto x11Smooth = !touch
```
ilya-fedin, line 291: Why the current capability check doesn't suffice?

ui/ui_utility.cpp
```diff
@@ -272,18 +274,38 @@ QPointF ScrollDeltaF(not_null<QWheelEvent*> e, bool touch) {
...
+		&& ::Platform::IsX11()
+		&& (pixelInput || !pixelDelta.isNull() || x11Touchpad);
+	if (x11Smooth && !angleDelta.isNull()) {
```
ilya-fedin, line 294: You're requiring both pixel delta and angle delta to be not null for the angle delta codepath?

ui/ui_utility.cpp
```diff
@@ -272,18 +274,38 @@ QPointF ScrollDeltaF(not_null<QWheelEvent*> e, bool touch) {
...
+		return (convert(angleDelta)
+			* QApplication::wheelScrollLines()
+			* kX11SmoothScrollMultiplier)
+			/ float64(QWheelEvent::DefaultDeltasPerStep);
+	}
+	if (pixelInput || (x11Smooth && !pixelDelta.isNull())) {
```
ilya-fedin, line 300: You already checked pixelDelta in the x11Smooth check?

ui/ui_utility.cpp
```diff
@@ -272,18 +274,38 @@ QPointF ScrollDeltaF(not_null<QWheelEvent*> e, bool touch) {
...
+		return (::Platform::IsX11() && !touch
+			? QPointF(pixelDelta)
```
ilya-fedin, line 302: You're breaking scrolling with scaling for X11 users

ui/widgets/elastic_scroll.cpp
```diff
@@ -867,7 +867,7 @@ bool ElasticScroll::eventHook(QEvent *e) {
 	switch (e->type()) {
 	case QEvent::ScrollPrepare: {
 		QScrollPrepareEvent *se = static_cast<QScrollPrepareEvent *>(e);
-		se->setViewportSize(QSizeF(viewport()->size()));
+		se->setViewportSize(QSizeF(size()));
```
ilya-fedin, line 870: Why this change?

ui/widgets/elastic_scroll.cpp
```diff
@@ -1029,21 +1029,29 @@ bool ElasticScroll::handleWheelEvent(not_null<QWheelEvent*> e, bool touch) {
 	const auto guard = gsl::finally([&] {
 		_lastScroll = now;
 	});
-	auto unmultiplied = ScrollDelta(e, touch);
+	if (phase == Qt::ScrollBegin) {
+		_wheelDeltaRemainder = {};
+	}
+	auto exactPixels = ScrollDeltaF(e, touch);
```
ilya-fedin, line 1035: Does it really help?

ui/ui_utility.cpp
catap: I had tested different, this one make it works acceptable enough.

catap:
And yes, it is really help. This patch is included into tdesktop in OpenBSD ports, and I do have some positive reports that scroll finally works. Not perfect, but better.

ilya-fedin:
> that scroll finally works

While currently it doesn't work at all?

catap:
Currently it works, but it is extreamly sensetive, and any scroll moves to many messages up/down and make usage nearly impossible.

Scroll in contancts, not chat, was broken for a while, but in chats it was broken recently, like a month or two.

ilya-fedin:
Can you answer each question in review? I'm not sure all changes are necessary and right.

ui/ui_utility.cpp
catap: Actually I use it with DPI ~270

catap:
@ilya-fedin I'll do it, but I need some time to recover context.

ui/ui_utility.cpp
ilya-fedin: I'm doubt you reach that line at all so I don't think you see the consequences of this change. Can you confirm you actually reach it?

nazarenkomykyta: > You're breaking scrolling with scaling for X11 users

You broke the whole scroll using ElasticScroll() on Xenocara, ironic...

ilya-fedin: Why? What's the problem with it on Xenocara?

nazarenkomykyta: > Why? What's the problem with it on Xenocara?

Do you understand the context of what people are saying to you? 

It's **completely broken** on Xenocara, without this patch, it's impossible to scroll normally.

ilya-fedin: > Do you understand the context of what people are saying to you?

No, I don't. I don't even know what Xenocara is (sounds like some X-on-X server?) and why ElasticScroll is broken on it.

> It's **completely broken** on Xenocara, without this patch, it's impossible to scroll normally.

This patch looks broken, I asked author to at least explain what each change does but he disappeared... If it's merged as it is, it will maybe fix scroll on Xenocara but will definitely break scrolling on standard Xorg server.

catap: @ilya-fedin I do not disappeared, as I said I need some time to recover context and explain it in details. It is not forgotten, I just overload differnet things.

Xenocara is a name for OpenBSD project which ships X11 (Xorg) and other X11 stuff.

I assume you tested it on Wayland because usual Linux uses it, and as far as I hear the same complains was from NetBSD world where X11/Xorg is used as well.

ui/widgets/elastic_scroll.cpp
catap: Yes, it does, and I fight here rounding by moving to float.

ui/ui_utility.cpp
catap: Sorry I don't understand you.

ui/ui_utility.cpp
catap: What is copy and paste issue and side effect from iterated development of that patch to actually fix the scroll, and you are right that || condition never reached.

Before I address it and possible other review point, I'd like to understand do you at least consider to merge that fix in some way?

ui/ui_utility.cpp
ilya-fedin: > I assume you tested it on Wayland because usual Linux uses it

The userbase is so high that a bug here will lead to a high stream of issues even from X11 users. In fact when ElasticScroll was just made, there were complaints from users of all platforms, including X11-specific ones. There were multiple rounds of fixes until complaints stopped.

ui/ui_utility.cpp
ilya-fedin: Before your change, it was checking for `e->device()->capabilities().testFlag(PixelScroll)`. This capability could be ever set only on a touchpad, it also replaces the `!pixelDelta.isNull()` check. I don't see what we get with an explicit check for a touchpad + `!pixelDelta.isNull()` but I see that it will lead to bugs on Wayland (it could return a null QPoint due to rounding which is compensated in the next frame).

ui/ui_utility.cpp
ilya-fedin: That's a very hard question since generally I'm not opposed to fixes but the current feeling from the patch is that it masks some bug in OpenBSD input driver and when we decompose the patch to the real fix it could be that we would need either of two:
1. A fix in OpenBSD code
2. An explicit check for OpenBSD in tdesktop code to work around the OpenBSD bug

But I could be wrong of course so I have no idea what the end result will be. I propose to check the behavior on last tdesktop version without any patch firstly since it feels the patch was rebased a lot of times and not checked whether it's (or at least all parts of it) still needed.

ui/widgets/elastic_scroll.cpp
ilya-fedin: You mean the same what was done in #324? Author then concluded that subpixel scroll makes things feel worse and left just the main thread offload in https://github.com/telegramdesktop/tdesktop/pull/31023 (https://github.com/telegramdesktop/tdesktop/commit/4bf51f6f5f49f521a8a76efa405dee6b9468175f is the commit from the older version of the PR to make use of #324)

ui/ui_utility.cpp
catap: Yes, this patch was origianlly made for tdesktop 7.0.6, and after that was "rebased" to current master and tested by 64gram's port which was made by @nazarenkomykyta and I think he used patch from that PR.

I not sure that it is unique OpenBSD bug, because, as I said, NetBSD comunity had shared same issues.

Also, before recent update of tdesktop port 6.9.3 -> 7.0.6 I had issue with scroll only on contact / chat list, and not in chat messages. I had it for I think like a year (or longer, not sure) but I wasn't motivated enough to dig it.

Also, original (unpatched) version hasn't got scroll issue out of the box. It needs many contacts / chats in folder and it works well with few of them. And the same with chat. When I created a new group chat, here no scroll issue, but after some contetn it started to appears, and more messages in chat makes it worse.

ui/ui_utility.cpp
catap: Well, it doesn't work on OpenBSD and NetBSD (to my knwoledge). But I think nobody reported it to you. So, I think assumption that here will be a lot of complains is not true.

ui/widgets/elastic_scroll.cpp
catap: Yes, and I'd like to say that current patch hasn't make scroll ideal. It is not that bad, but not ideal.

And, frankly, ElasticScroll logic is very complicated / tricky, so, yes, I can broke something somewhere and probably it can be made better.

Which input do you need to make sane investigation from scratch?

ui/ui_utility.cpp
ilya-fedin: I was talking solely about Linux, that it couldn't be broken on Linux with X11 and there be no mass reporting. I don't doubt that there's not much users on BSDs to get reports, especially given the lack of official builds.

ui/ui_utility.cpp
ilya-fedin: I don't know much about BSDs but I assume OpenBSD and NetBSD could share the same driver code?

ui/widgets/elastic_scroll.cpp
ilya-fedin: Personally I have no idea what could go wrong on OpenBSD. But I tried to feed your PR to Claude and it confirmed my suspicions about driver bug in OpenBSD, even proposed a patch to it (`xenocara/driver/xf86-input-ws/src/ws.c`):
```
-	SetScrollValuator(pWS, HSCROLL_AXIS, SCROLL_TYPE_HORIZONTAL, 4096, 0);
-	SetScrollValuator(pWS, VSCROLL_AXIS, SCROLL_TYPE_VERTICAL, 4096, 0);
+	SetScrollValuator(pWS, HSCROLL_AXIS, SCROLL_TYPE_HORIZONTAL, 120, 0);
+	SetScrollValuator(pWS, VSCROLL_AXIS, SCROLL_TYPE_VERTICAL, 120, 0);

 	if (hw.hscroll || hw.vscroll) {
 		valuator_mask_zero(priv->scroll_mask);
 		valuator_mask_set_double(priv->scroll_mask,
-		    HSCROLL_AXIS, (double) hw.hscroll);
+		    HSCROLL_AXIS, hw.hscroll * 120.0 / 4096.0);
 		valuator_mask_set_double(priv->scroll_mask,
-		    VSCROLL_AXIS, (double) hw.vscroll);
+		    VSCROLL_AXIS, hw.vscroll * 120.0 / 4096.0);
```
But it also said that the current tdesktop version shouldn't be affected by it (7.0.6 still does).

ui/ui_utility.cpp
catap: Some? Yes. Probably https://man.openbsd.org/wscons.4 and friends is more or less similar, but I have no idea how similarity has.

ui/ui_utility.cpp
catap: Well, I think majority of Linux users are moved to Wayland which default options at Fedora, Ubuntu and so on.

ui/widgets/elastic_scroll.cpp
catap: I had tried 64gram version 1.2.7 which is based on 7.0.9 tdesktop and it had the same issue.

Regarding your patch, I have no knowledge of that code at all and I'd like to ask @mherrb, maybe he had time to help us.

ui/ui_utility.cpp
ilya-fedin: There's still enough Linux X11 users. You can see an issue opened by Xwayland user just 5 hours ago and two PRs opened yesterday from a real X11 user (not Xwayland), one of them already merged though. And then periodically I see reports from users on ancient Debian/Devuan and X11 of course.

ui/widgets/elastic_scroll.cpp
ilya-fedin: You need 7.1.0 or later

catap: Unfortently OpenBSD's fido2 library is too old to build it https://github.com/openbsd/src/tree/master/lib/libfido2

It fails as:
```
/usr/ports/pobj/tdesktop-7.1.3/tdesktop-7.1.3-full/Telegram/SourceFiles/webauthn/webauthn_common.cpp:356:6: error: use of undeclared identifier 'fido_assert_authdata_raw_ptr'; did you mean 'fido_assert_authdata_ptr'?
  356 |                                         fido_assert_authdata_raw_ptr(assert, 0)),
      |                                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                         fido_assert_authdata_ptr
/usr/include/fido.h:61:22: note: 'fido_assert_authdata_ptr' declared here
   61 | const unsigned char *fido_assert_authdata_ptr(const fido_assert_t *, size_t);
      |                      ^
/usr/ports/pobj/tdesktop-7.1.3/tdesktop-7.1.3-full/Telegram/SourceFiles/webauthn/webauthn_common.cpp:357:5: error: use of undeclared identifier 'fido_assert_authdata_raw_len'; did you mean 'fido_assert_authdata_len'?
  357 |                                 fido_assert_authdata_raw_len(assert, 0));
      |                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                 fido_assert_authdata_len
/usr/include/fido.h:171:8: note: 'fido_assert_authdata_len' declared here
  171 | size_t fido_assert_authdata_len(const fido_assert_t *, size_t);
      |        ^
2 errors generated.

```

I'll try to cook something to test in next days.

ilya-fedin: Claude suggested that you can try to just add
```
elseif (CMAKE_SYSTEM_NAME MATCHES "OpenBSD")
    list(APPEND fido2_sources
        ${fido2_src}/hid_openbsd.c
        ${fido2_src}/hid_unix.c)
```
and remove udev link in lib_fido2.cmake and use the bundled one

nazarenkomykyta: > Claude suggested that you can try to just add
> 
> ```
> elseif (CMAKE_SYSTEM_NAME MATCHES "OpenBSD")
>     list(APPEND fido2_sources
>         ${fido2_src}/hid_openbsd.c
>         ${fido2_src}/hid_unix.c)
> ```
> 
> and remove udev link in lib_fido2.cmake and use the bundled one.

nazarenko$ cat patch-Telegram_SourceFiles_webauthn_webauthn_common_cpp  
```
Replace `raw_ptr` and `raw_len` with `_ptr` and `_len` to ensure
compatibility with lib_fido2. Otherwise, the 64gram will fail to
compile.

--- Telegram/SourceFiles/webauthn/webauthn_common.cpp.orig      Wed Aug  5 16:13:06 2026
+++ Telegram/SourceFiles/webauthn/webauthn_common.cpp   Wed Aug  5 16:15:54 2026
@@ -320,8 +320,8 @@
                                fido_assert_id_len(assert, 0));
                        out.authenticatorData = QByteArray(
                                reinterpret_cast<const char*>(
-                                       fido_assert_authdata_raw_ptr(assert, 0)),
-                               fido_assert_authdata_raw_len(assert, 0));
+                                       fido_assert_authdata_ptr(assert, 0)),
+                               fido_assert_authdata_len(assert, 0));
                        out.signature = QByteArray(
                                reinterpret_cast<const char*>(fido_assert_sig_ptr(assert, 0)),
                                fido_assert_sig_len(assert, 0));
```

And that's all.

ilya-fedin: Maybe it compiles but fido won't actually work that way....

nazarenkomykyta: > Maybe it compiles but fido won't actually work that way....

https://developers.yubico.com/libfido2/Manuals/fido_assert_authdata_raw_ptr.html

`The fido_assert_authdata_ptr(), fido_assert_authdata_raw_ptr(), fido_assert_clientdata_hash_ptr(), fido_assert_id_ptr(), fido_assert_user_id_ptr(), fido_assert_sig_ptr(), fido_assert_sigcount(), and fido_assert_flags() functions return pointers to the CBOR-encoded and raw authenticator data, client data hash, credential ID, user ID, signature, signature count, and authenticator data flags of statement idx in assert.`

I've been using this hack for two weeks now, and it seems to be working.

ilya-fedin: > I've been using this hack for two weeks now, and it seems to be working.

This code path is for USB keys. Did you log in with a USB key?

nazarenkomykyta: > Otherwise, the 64gram will fail to
> compile.

> This code path is for USB keys. Did you log in with a USB key?

No, and I don't think it makes any sense on OpenBSD anyway. 

What's great about this hack is that you don't have to add workarounds to the build system.

catap: @ilya-fedin for test of scroll I don't need actually working fido, so, I just used patch from @nazarenkomykyta , I have no idea does it broke fido or not, but it allows to compile tdesktop-7.1.3

So far scroll is very sensitive, but at least it can be used.

I'll keep your posted.

catap:
Ok, the last version of tdesktop works not perfect but works, and because I also have no time or motivation to dig more into scroll... and because @nazarenkomykyta looks like a man who willing to dig into it, well, I grand access to my fork should allow him to adjust that branch and do not reopen PR to continue discussion in this place.

Alternativly this PR can be closed.

Both way fine for me, and I sent to ports@openbsd.org update of tdesktop and remove of my patches: https://marc.info/?l=openbsd-ports&m=178799901300513&w=2

ilya-fedin:
I assume silence means a new PR will be opened if there's still anyone is interested in the work

nazarenkomykyta:
@ilya-fedin 

I also tested scrolling on Wayland ( Hyprland ), and the touchpad sensitivity is extremely high, just like on Xenocara. I think I'll open a new pull request as soon as I come up with a solution.

PR #342, opened 2026-08-20
Fix SeparatePanel fullscreen origin on Windows

anriltine:
## Summary
- After `showFullScreen()`, explicitly move/resize the frameless `SeparatePanel` to the current screen geometry.
- Fixes Mini Apps that were resized for fullscreen but stayed anchored at the pre-fullscreen panel position on Windows (often with multiple monitors).

## Context
Reported in [telegramdesktop/tdesktop#30963](https://github.com/telegramdesktop/tdesktop/issues/30963) (also referenced by #31041 / #31114).

Root cause matches QTBUG-39537 / QTBUG-86899 for frameless windows: `QWidget::showFullScreen()` applies the fullscreen size but drops the move to the screen origin. Media viewer already works around this by setting screen geometry explicitly; Mini Apps use `SeparatePanel`, which previously only called plain `showFullScreen()`.

## Test plan
- [ ] Windows, single monitor: open a Mini App, enter fullscreen via UI or `web_app_request_fullscreen` — panel should cover the full screen from the monitor origin.
- [ ] Windows, two monitors: main Telegram window mid-primary; open Mini App fullscreen — panel should cover the primary screen, not remain centered over the main window and overflow.
- [ ] Exit fullscreen — panel returns to the normal anchored size/position.
- [ ] Linux/macOS smoke: Mini App fullscreen still enters/exits correctly.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/ui/widgets/separate_panel.cpp
+++ b/ui/widgets/separate_panel.cpp
@@ -1388,6 +1388,17 @@ void SeparatePanel::toggleFullScreen(bool fullscreen) {
 	_fullscreen = fullscreen;
 	if (fullscreen) {
 		showFullScreen();
+		// On Windows, frameless showFullScreen() may resize without moving to
+		// the screen origin (QTBUG-39537). Mini apps then stay anchored at the
+		// pre-fullscreen panel position - see telegramdesktop/tdesktop#30963.
+		if (const auto current = screen()
+				? screen()
+				: QGuiApplication::primaryScreen()) {
+			const auto geometry = current->geometry();
+			if (this->geometry() != geometry) {
+				Ui::SetGeometryAndScreen(this, geometry);
+			}
+		}
 	} else {
 		showNormal();
 	}
--- a/ui/widgets/separate_panel.cpp
+++ b/ui/widgets/separate_panel.cpp
@@ -1387,18 +1387,14 @@ QRect SeparatePanel::innerGeometry() const {
 void SeparatePanel::toggleFullScreen(bool fullscreen) {
 	_fullscreen = fullscreen;
 	if (fullscreen) {
+		// RpWidget's constructor calls setGeometry(0, 0, 0, 0), which sets
+		// WA_Moved. After that Qt treats the panel as explicitly positioned
+		// and QWidget::show() / QWidgetPrivate::create pass that position to
+		// the QWindow instead of letting the platform place a fullscreen
+		// window at the screen origin (resize-only). Same workaround as the
+		// Linux transient-parent path in initGeometry().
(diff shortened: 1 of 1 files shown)


ui/widgets/separate_panel.cpp
```diff
@@ -1388,6 +1388,17 @@ void SeparatePanel::toggleFullScreen(bool fullscreen) {
 	_fullscreen = fullscreen;
 	if (fullscreen) {
 		showFullScreen();
+		// On Windows, frameless showFullScreen() may resize without moving to
+		// the screen origin (QTBUG-39537). Mini apps then stay anchored at the
```
ilya-fedin: You're referencing a fixed Qt bug

ilya-fedin:
> Root cause matches QTBUG-39537 / QTBUG-86899 for frameless windows

QTBUG-39357 is fixed long time ago while QTBUG-86899 seem to be completely unrelated (it claims maximized windows behave like fullscreen ones on non-primary monitors and hide taskbar which doesn't seem to be the problem here)?

ilya-fedin:
The root cause could be not a Qt bug at all but our bug: Qt positions windows automatically only unless setGeometry() or move() are called, RpWidget initializes with zero geometry. The Linux transient parent code path already works around that by setting WA_Moved attribute to false, maybe the same should be done before showFullscreen().

anriltine:
Thanks, that matches what the Qt sources do.

`QTBUG-39537` / `QTBUG-86899` was the wrong diagnosis: current Qt already moves a fullscreen window to the monitor rect in `QWindowsWindow::setWindowState_sys()`. The remaining failure is on our side. `RpWidget` starts with `setGeometry(0, 0, 0, 0)`, which sets `WA_Moved`, so `QWidgetPrivate::create` / `QWidget::show()` pass that position into the `QWindow` (`setGeometry`) instead of resize-only placement at the screen origin.

The previous patch also fought Qt: `SetGeometryAndScreen()` after `showFullScreen()` goes through `setGeometry_sys()`, which clears `Qt::WindowFullScreen` when not inside `setWindowState()`.

Updated to the same workaround as the Linux transient-parent path in `initGeometry()`: `setAttribute(Qt::WA_Moved, false)` before `showFullScreen()`.

anriltine:
@ilya-fedin thanks again for the review. #342 was the wrong approach (QTBUG citations + `SetGeometryAndScreen()` after `showFullScreen()`).

Opened a clean PR with the `WA_Moved` fix you suggested, based on current master: #343

anriltine:
Superseded by #343.

PR #345, opened 2026-08-24
Add a selection event for a painted list child

rezabakhshilaktasaraei:
A painted list has no way to tell a screen reader that a row became the current one, or stopped being. accessibilityChildStateChanged() with the selected flag looks like it should do it, but on Windows the bridge reads only the checked flag out of a state change (qwindowsuiamainprovider.cpp, notifyStateChange). So the chat folder tabs stayed silent on activation, while the folders that are real widgets say "selected" - SideBarButton::setActive() raises the selection event itself.

accessibilityChildSelectionChanged(index) does the same for a virtual child. It raises SelectionAdd for a deselection as well, which reads odd but is deliberate: SelectionRemove is mapped nowhere in the bridge (in qtbase dev either, not just 5.15), and a screen reader reads the new value off the child and announces the difference - so it only needs telling that the selection of this child changed. Verified in a log: NVDA answers the event with a state comparison, saying just "selected" or "not selected", and stays quiet when the value did not actually change.

Used by telegramdesktop/tdesktop#31097 when a slider section is activated, and by telegramdesktop/tdesktop#31171 when a message is selected or deselected. Both verified with NVDA on Windows 10.

### DIFF
--- a/ui/rp_widget.cpp
+++ b/ui/rp_widget.cpp
@@ -457,6 +457,12 @@ void RpWidget::accessibilityChildStateChanged(
 	QAccessible::updateAccessibility(&event);
 }
 
+void RpWidget::accessibilityChildSelectionChanged(int index) {
+	auto event = QAccessibleEvent(this, QAccessible::SelectionAdd);
+	event.setChild(index);
+	QAccessible::updateAccessibility(&event);
+}
+
 void RpWidget::accessibilityChildFocused(int index) {
 	QAccessibleEvent event(this, QAccessible::Focus);
 	event.setChild(index);
--- a/ui/rp_widget.h
+++ b/ui/rp_widget.h
@@ -457,6 +457,16 @@ public:
 	void accessibilityChildValueChanged(int index);
 	[[nodiscard]] virtual QAccessible::State accessibilityChildState(int index) const;
 	void accessibilityChildStateChanged(int index, AccessibilityState changes);
+
+	// Announces that a child was selected or deselected. A state change does
+	// not cover it: the Windows bridge looks at the checked flag there and
+	// ignores the selected one. Both directions raise the same event, the one
+	// that bridge forwards (SelectionRemove reaches no one) - a screen reader
+	// reads the new value off the child and announces the difference, so it
+	// only needs telling that the selection changed. SideBarButton does the
+	// same for the folders that are real widgets.
+	void accessibilityChildSelectionChanged(int index);
+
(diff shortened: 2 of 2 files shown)


john-preston:
**Verdict: CHANGES_REQUIRED.** Preserve the small helper, but preserve selection direction and correct the Windows bridge mapping before treating it as a general selection-change notification. The finding is a confirmed event-contract mismatch, not a reproduced screen-reader announcement failure.

**P2 — Preserve the meaning of deselection notifications.** [ui/rp_widget.cpp:461](https://github.com/desktop-app/lib_ui/blob/3656218feae9b6a7a26155ac8f6148145faebb60/ui/rp_widget.cpp#L460-L464) unconditionally emits `SelectionAdd`. With tdesktop#31171, select a message and press Space again, or shrink a Shift+arrow range: the helper runs after the message's selected state becomes false. With #31097, explicit tab activation also calls it for the previously selected tab. Qt's Windows dispatcher forwards `SelectionAdd` to a provider that unconditionally raises `UIA_SelectionItem_ElementSelectedEventId`; it neither checks the new selected state nor emits the removal event. This is present in [Qt 5.15.2's dispatcher](https://github.com/qt/qtbase/blob/v5.15.2/src/plugins/platforms/windows/uiautomation/qwindowsuiaaccessibility.cpp#L173-L175) and [Qt 6.11.0's provider](https://github.com/qt/qtbase/blob/v6.11.0/src/plugins/platforms/windows/uiautomation/qwindowsuiamainprovider.cpp#L192-L198). Microsoft's [UIA event contract](https://learn.microsoft.com/en-us/windows/win32/winauto/uiauto-event-ids) distinguishes an item removed from the selection from the single selected item. Consequently this helper sends a selection assertion on an unselected object. The explicit selection-event path supplies no removal; however, configured project Qt 5 already receives correct removal events from preceding selected-state changes, as clarified below. The submitted description reports that NVDA compensates by re-reading state; it does not establish equivalent behavior for other clients. No false spoken output was reproduced in this review.

Choose a coherent per-backend notification strategy. Merely changing Add to Remove leaves the explicit Windows path silent, although configured Qt 5 already has selected-state forwarding. If keeping explicit events, derive direction from current selected state, repair their Windows mapping and avoid duplicating the existing state-driven notifications. A shared notification helper can centralize this instead of adding another state cache. Preserve required state events for Cocoa and explicit selection coverage for AT-SPI; their backend paths differ.

PR #354, opened 2026-08-27
Expose a numeric range value and radio selection items

rezabakhshilaktasaraei:
Two additions the value-picking sliders in tdesktop need:

- RpWidget can report a numeric range (accessibilityValueRange) and the accessible wrapper then exposes QAccessibleValueInterface, which UI Automation maps to the RangeValue pattern. Value change events keep carrying the textual value - that is the one a screen reader announces and reads the value back from.
- RadioButton is accepted as a selection item role next to ListItem and PageTab, so a group of painted radio buttons resolves its checked item through the selection interface.

Built on top of #348 - only the last two commits are new.

### DIFF
--- a/ui/abstract_button.h
+++ b/ui/abstract_button.h
@@ -69,9 +69,17 @@ public:
 	[[nodiscard]] bool isListItem() const {
 		return _listItem;
 	}
+	void setIsPageTab(bool value) {
+		_pageTab = value;
+	}
+	[[nodiscard]] bool isPageTab() const {
+		return _pageTab;
+	}
 
 	QAccessible::Role accessibilityRole() override {
-		return _listItem
+		return _pageTab
+			? QAccessible::PageTab
+			: _listItem
 			? QAccessible::ListItem
 			: _menuButton
 			? QAccessible::ButtonMenu
@@ -132,6 +140,7 @@ private:
 	bool _triggerOnPress : 1 = false;
 	bool _menuButton : 1 = false;
 	bool _listItem : 1 = false;
+	bool _pageTab : 1 = false;
 
 	Fn<void()> _clickedCallback;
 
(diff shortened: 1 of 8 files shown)


john-preston:
**Verdict: CHANGES_REQUIRED.** Keep the small opt-in value adapter and radio-role extension. Complete numeric change notifications for the newly advertised RangeValue pattern.

**P2 — Notify the numeric range property when its value changes.** At [ui/rp_widget.cpp:523-527](https://github.com/desktop-app/lib_ui/blob/0948a783002a9add06f643b3dab99fb19bf5ddf7/ui/rp_widget.cpp#L523-L527), the existing helper still constructs its event from QString. With #31191, focus the notification-count slider and change 2 to 3 with an arrow or RangeValue.SetValue: [changeValueTo](https://github.com/telegramdesktop/tdesktop/blob/9d27388746da84cf39434d96084f0c154d41bc76/Telegram/SourceFiles/ui/widgets/discrete_sliders.cpp#L376-L381) invokes only this helper. Both the configured Qt 5.15.19 backend and [Qt 6.11.2](https://github.com/qt/qtbase/blob/v6.11.2/src/plugins/platforms/windows/uiautomation/qwindowsuiamainprovider.cpp#L134-L148) route a QString exclusively to UIA_ValueValuePropertyId; the numeric RangeValue event is in the alternate branch. Consequently a client subscribed to RangeValue.Value gets no change notification even though querying the property returns the new number. Microsoft's [slider requirements](https://learn.microsoft.com/en-us/windows/win32/winauto/uiauto-supportslidercontroltype#required-events) require that event when RangeValue is supported. This is a source-confirmed event-contract omission, not a reproduced speech failure.

Make range-bearing controls issue a numeric value-change notification while preserving textual notifications where clients require them. Keep this policy in one range-aware helper or a bridge fan-out, rather than duplicating it in every consumer. Validate that supporting both Windows properties does not produce duplicate speech or duplicate AT-SPI/Cocoa notifications. The configured AT-SPI and Cocoa backends do not have the same string-versus-numeric branch, so this finding is specifically the Windows contract.

PR #356, opened 2026-08-29
Accessibility: auto-fallback button accessible names

hesam-oxe:
## Summary

`Ui::AbstractButton` now exposes an accessible name automatically for screen readers (NVDA/JAWS/ORCA):

- If an explicit `setAccessibleName()` was set, it is used as before.
- Otherwise the button's `toolTip()` is used as a fallback.

`FlatButton` and `SettingsButton` also call `setAccessibleName()` from their text.

This means icon-only buttons (close, mute, record, play, etc.) that previously had no accessible name now announce their tooltip, addressing the missing-name reports in telegramdesktop/tdesktop#476. It is the foundational, low-risk part of a larger accessibility effort; focus-navigation, context-menu focus announcements, inline-media `InvokePattern`, and voice-recording keyboard alternatives remain follow-up work.

## Changes

- `ui/abstract_button.cpp` / `ui/abstract_button.h`: added `AbstractButton::accessibilityName()` override with the explicit-name -> tooltip fallback.
- `ui/widgets/buttons.cpp`: `FlatButton::setText` and `SettingsButton::setText` set the accessible name from their text.

## Related

- Consumed by telegramdesktop/tdesktop#476 (PR on `feat/accessibility-base-refactor`).
- Issue: telegramdesktop/tdesktop#476


ui/abstract_button.cpp
```diff
@@ -223,6 +223,18 @@ void AbstractButton::clearState() {
...
+QString AbstractButton::accessibilityName() {
+	const auto custom = QWidget::accessibleName();
+	if (!custom.isEmpty()) {
+		return custom;
+	}
+	const auto tip = toolTip();
+	if (!tip.isEmpty()) {
+		return tip;
+	}
+	return QString();
```
ilya-fedin, line 235: This is unnecessary complicated.
```suggestion
	return toolTip();
```

ui/widgets/buttons.cpp
```diff
@@ -239,6 +239,7 @@ FlatButton::FlatButton(
 
 void FlatButton::setText(const QString &text) {
 	_text = text;
+	setAccessibleName(text);
```
ilya-fedin, line 242: `FlatButton::accessibilityName` already returns `_text`, this change has no sense

ui/widgets/buttons.cpp
```diff
@@ -1056,6 +1057,7 @@ void SettingsButton::onStateChanged(
 
 void SettingsButton::setText(TextWithEntities &&text) {
 	_text.setMarkedText(_st.style, text, kMarkupTextOptions, _context);
+	setAccessibleName(_text.toString());
```
ilya-fedin, line 1060: `SettingsButton::accessibilityName` already returns `_text.toString()`, this change has no sense

ui/abstract_button.cpp
hesam-oxe: Done in 5511f33 — simplified to `return toolTip();`. You are right that the explicit name check was redundant: `Widget::text(QAccessible::Name)` in `ui_accessible_widget.cpp` already returns `QAccessibleWidget::text()` first, so an explicitly set `accessibleName` never reaches this override. Thanks for pointing that out.

ui/widgets/buttons.cpp
hesam-oxe: Done in 5511f33 — removed the `setAccessibleName(text)` call; `FlatButton::accessibilityName()` already returns `_text`, so the property set was indeed pointless.

ui/widgets/buttons.cpp
hesam-oxe: Done in 5511f33 — removed here as well; `SettingsButton::accessibilityName()` already returns `_text.toString()`.

ilya-fedin: please integrate the fixes into the commit introducing the problems rather than leaving problems in commit history

hesam-oxe:
@cla-assistant check

hesam-oxe:
@ilya-fedin Done — the review fixes are now folded into the commit that introduced the problems. The branch is now two commits:

- `a426078` **Accessibility: auto-fallback button accessible names** — now directly contains the simplified `return toolTip();` and no longer adds the redundant `setAccessibleName()` calls on `FlatButton`/`SettingsButton` (your three review points are addressed in this commit itself, not in a follow-up).
- `3e962ef` **Accessibility: expose InvokePattern on message sub-items** — unchanged content.

The net diff against the previous head is empty — only the history shape changed, so nothing new to review. Sorry it took an extra round to get the history right.

hesam-oxe:
@cla-assistant check

hesam-oxe:
@ilya-fedin Following up on your note in #357 that "#356 still contains the bugged commit" — I think our messages crossed, so let me pin the current state of this branch down explicitly.

The history rewrite you asked for ("integrate the fixes into the commit introducing the problems rather than leaving problems in commit history") is pushed. The branch is now exactly two commits:

- `a426078` **Accessibility: auto-fallback button accessible names** — contains the simplified `return toolTip();` in `ui/abstract_button.cpp` directly, and no longer adds the redundant `setAccessibleName()` calls on `FlatButton` (`ui/widgets/buttons.cpp:242`) or `SettingsButton` (`ui/widgets/buttons.cpp:1060`). All three of your review points are addressed *inside* the commit that introduced them, not in a follow-up commit.
- `3e962ef` **Accessibility: expose InvokePattern on message sub-items** — content unchanged (`ui/accessible/ui_accessible_item.*`, `ui/rp_widget.*`).

The net diff against the previous head is empty; only the history shape changed, so there is nothing new to review beyond confirming the squash is the shape you wanted.

Separately: I've closed #357 as redundant per your suggestion — master already fires the per-item `QAccessible::Focus` event via `ItemBase::setSelected()`, so the extra `PopupMenu`-level event had nothing to add.

ui/widgets/buttons.cpp
hesam-oxe: Done — the review fixes are now folded into the introducing commit (`24074c5`) rather than living as a separate fixup commit. The PR is two commits: `24074c5` carries the simplified `AbstractButton::accessibilityName() { return toolTip(); }` (no explicit-name branch, no setAccessibleName calls in FlatButton/SettingsButton), and `d9ff122` is the InvokePattern work for message sub-items. I also dropped the stale last sentence from the first commit message that still described the removed setAccessibleName approach.

ilya-fedin:
> It is the foundational, low-risk part of a larger accessibility effort; focus-navigation, context-menu focus announcements, inline-media `InvokePattern`, and voice-recording keyboard alternatives remain follow-up work.

Do you think it's a good idea proposing 3 lines accessibility name change in the same PR with a work that adds new feature and new overloads to the core class of all widgets in the app?

hesam-oxe:
@ilya-fedin You're right — splitting as suggested. Prepared two clean branches from current upstream master (no rewrite of this PR yet):

- Branch A (narrow, 3-line fallback only): hesam-oxe/lib_ui feat/button-accessible-names (79fc402) — ui/abstract_button.cpp/h only, no InvokePattern/overloads.
- Branch B (feature): hesam-oxe/lib_ui feat/invoke-pattern-sub-items (b2dc0f3) — ui/accessible/ui_accessible_item.*, ui/rp_widget.* only.

This PR is left untouched pending your confirmation (avoiding a force-push until you confirm the split shape). Once you confirm, I'll retarget this PR to Branch A content and open the Branch B PR, cross-linking both. Sorry it took a couple of rounds to land on this shape.

hesam-oxe:
@ilya-fedin Split as you suggested:
- PR #369: button accessible names (3 lines)
- PR #370: InvokePattern on message sub-items
Thanks for the guidance.

PR #359, opened 2026-09-01
Report the position of a painted list item within its set

rezabakhshilaktasaraei:
A painted list item had no way to tell a screen reader which of how many rows it is, so the lists that are real widgets say "2 of 10" while the painted ones stay silent about it.

Item reports it now through the accessibility attributes interface, as the PositionInSet / SizeOfSet attributes the Windows UIA bridge maps to the properties a screen reader announces. The numbers come from the owner through accessibilityChildSetPosition(): by default every child is one of the set (index + 1 of the child count), and a list with rows that are not items - a divider between them - overrides it to leave those out, so they neither get a position nor shift the ones below. A row outside the set does not expose the interface at all.

The two attributes are in Qt since 6.13 (https://codereview.qt-project.org/c/qt/qtbase/+/764810), so the Qt 5.15 and 6.11.1 builds need desktop-app/patches#265 to compile.

Used by telegramdesktop/tdesktop#31218 for the message lists. Verified with NVDA on Windows 10: the chat list, the folder tabs, the countries and languages lists and the message lists all announce "n of N" after the row name.

### DIFF
--- a/ui/accessible/ui_accessible_item.cpp
+++ b/ui/accessible/ui_accessible_item.cpp
@@ -198,18 +198,47 @@ QAccessibleInterface *Item::parent() const {
 }
 
 void *Item::interface_cast(QAccessible::InterfaceType type) {
+	const auto parent = _parent.get();
+	const auto index = parent ? currentIndex() : -1;
 	if (type == QAccessible::ActionInterface) {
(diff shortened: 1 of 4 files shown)


john-preston:
Please, resolve conflicts.

PR #360, opened 2026-09-02
Fix the keyboard handling of a radio group

rezabakhshilaktasaraei:
Two fixes for the keyboard handling of a radio group, both in Radiobutton.

The roving Tab stop of the group was given Qt::StrongFocus, which also takes keyboard focus on a mouse click - a change for everyone, while all the group needs is to be reachable with Tab. Qt::TabFocus is what the rest of the accessibility code resolves its focusable widgets to, and it leaves the mouse behaviour alone.

The arrows move between the buttons along the axis the group is laid out on, found by comparing the position of a button with the one next to it. Those positions were read inside their parents, which works only while the buttons are siblings. The correct answer buttons of a quiz (Ui::FadeWrapScaled in create_poll_box), the rows of Settings > Privacy > Auto-delete messages and the premium gift options each put their button in a wrap of its own, at the same position inside it, so the group looked horizontal: Up / Down did nothing on a list of answers going down the box, while Left / Right moved through it. Comparing global positions fixes those and changes nothing where the buttons are siblings.

Verified with NVDA on Windows 10: in the quiz answers of a new poll the arrows now follow the list downwards and Left / Right are ignored, and the radio list of the self-destruct box, whose buttons are siblings, is unchanged.

### DIFF
--- a/ui/widgets/checkbox.cpp
+++ b/ui/widgets/checkbox.cpp
@@ -1036,14 +1036,14 @@ void Radiobutton::trackScreenReaderState() {
 		if (!screenReaderActive || !value.has_value()) {
 			return Qt::NoFocus;
 		} else if (value == _value) {
-			return Qt::StrongFocus;
+			return Qt::TabFocus;
 		}
 		for (const auto &button : _group->_buttons) {
 			if (button->_value == value) {
 				return Qt::NoFocus;
 			}
 		}
-		return Qt::StrongFocus;
+		return Qt::TabFocus;
 	}) | rpl::on_next([=](Qt::FocusPolicy value) {
 		if (focusPolicy() != value) {
 			setFocusPolicy(value);
--- a/ui/widgets/checkbox.cpp
+++ b/ui/widgets/checkbox.cpp
@@ -1092,8 +1092,14 @@ void Radiobutton::keyPressEvent(QKeyEvent *e) {
 		? buttons[currentIndex - 1]
 		: buttons[currentIndex + 1];
 
-	const auto deltaY = std::abs(neighbor->y() - y());
-	const auto deltaX = std::abs(neighbor->x() - x());
+	// In global coordinates: the buttons of one group are not always
+	// siblings - each one may sit in a wrap of its own, like the correct
+	// answer buttons of a quiz - and their positions inside those parents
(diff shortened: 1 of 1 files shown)


john-preston:
**Verdict: CHANGES_REQUIRED.** Keep this small implementation and add the control's own disabled flag to the eligibility check. The disabled-state finding is an incomplete fix of existing behavior, not a newly introduced ordinary-UX regression.

## Findings

1. **P2 — Also skip the radio's custom disabled state.** [checkbox.cpp:1121](https://github.com/desktop-app/lib_ui/blob/e429781f8d8c2761ad3d2c05b1f2eda4b8d05c78/ui/widgets/checkbox.cpp#L1121)

`Radiobutton` inherits [`AbstractButton::setDisabled()`](https://github.com/desktop-app/lib_ui/blob/e429781f8d8c2761ad3d2c05b1f2eda4b8d05c78/ui/abstract_button.cpp#L209), which changes `StateFlag::Disabled` without calling `QWidget::setEnabled()`. Its [`isDisabled()`](https://github.com/desktop-app/lib_ui/blob/e429781f8d8c2761ad3d2c05b1f2eda4b8d05c78/ui/abstract_button.h#L31) reads that separate flag. Consequently, with two visible vertically arranged radios A and B, A selected/focused and `B->setDisabled(true)`, Down still passes the new `isEnabled()` test, selects B, and focuses it. With an enabled C after B, navigation stops at B instead of skipping it.

This follows directly from the source; it was not run. The old arrow handler also selected B, so this is an incomplete disabled-choice fix. Pinned application callers do use this API for [restriction-duration radios](https://github.com/telegramdesktop/tdesktop/blob/86262333a457f62709726ee4ab9c48fa58824da4/Telegram/SourceFiles/boxes/peers/edit_participant_box.cpp#L933) and [unavailable call-device radios](https://github.com/telegramdesktop/tdesktop/blob/86262333a457f62709726ee4ab9c48fa58824da4/Telegram/SourceFiles/settings/settings_calls.cpp#L526), but those examples isolate the disabled choice; they do not demonstrate an existing mixed-group user flow. The minimum correction is `isVisible() && isEnabled() && !isDisabled()`, retaining the QWidget check for disabled ancestors and explicit QWidget disabling.

PR #362, opened 2026-09-03
Name a checkbox without a label by its accessible name

rezabakhshilaktasaraei:
`Checkbox::accessibilityName()` returns the label text, so a checkbox or radio button created without one — the correct answer button beside a poll answer, where the answer stands for the label — had no name for a screen reader and no way to be given one. Fall back to the widget's accessible name when the label is empty. Labelled checkboxes are unchanged.

### DIFF
--- a/ui/widgets/checkbox.h
+++ b/ui/widgets/checkbox.h
@@ -209,7 +209,11 @@ public:
 		return QAccessible::Role::CheckBox;
 	}
 	QString accessibilityName() override {
-		return _text.toString();
+		// A checkbox with no label of its own - beside a poll answer, in a
+		// list row - is named by what stands for it, through the widget's
+		// accessible name.
+		const auto text = _text.toString();
+		return text.isEmpty() ? RpWidget::accessibilityName() : text;
 	}
 	AccessibilityState accessibilityState() const override;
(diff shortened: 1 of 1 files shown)


ui/widgets/checkbox.h
```diff
@@ -209,7 +209,11 @@ class Checkbox : public RippleButton, public ClickHandlerHost {
 		return QAccessible::Role::CheckBox;
 	}
 	QString accessibilityName() override {
-		return _text.toString();
+		// A checkbox with no label of its own - beside a poll answer, in a
+		// list row - is named by what stands for it, through the widget's
+		// accessible name.
+		const auto text = _text.toString();
+		return text.isEmpty() ? RpWidget::accessibilityName() : text;
```
ilya-fedin, line 216: This is wrong, accessible name always takes the priority over this method

rezabakhshilaktasaraei:
You're right, QAccessibleWidget::text(Name) already returns the accessible name before Checkbox::accessibilityName() is consulted, so the fallback can never change the result. Verified with NVDA that the poll box radios are named without it; the poll box sets the name on the radio directly and that path is enough. Closing.

PR #363, opened 2026-09-03
Open a dropdown menu for a screen reader like a popup one

rezabakhshilaktasaraei:
A `PopupMenu` shown in screen reader mode opens as if from the keyboard: the first item is selected, takes the focus and is announced, arrows go on from there. A `DropdownMenu` (the attach menu of the composer, "send as") just appeared inside the window with nothing selected, so a screen reader had nothing to say and Space on the paperclip seemed to do nothing.

- `InnerDropdown` gets a shown callback, fired once the content is visible (after the show animation, or at once for `showFast`), mirroring the hidden callback.
- `DropdownMenu` uses it to select the first item in screen reader mode, like `PopupMenu` does.
- A selected menu item forces its lazy accessibility registration, so it can take the focus even when the screen reader has not queried it yet (a popup's new window gets queried, a menu inside the window may not).

Focus returns to the toggle on hide through the existing save/return logic. Mouse behaviour is unchanged.

### DIFF
--- a/ui/widgets/dropdown_menu.cpp
+++ b/ui/widgets/dropdown_menu.cpp
@@ -6,6 +6,8 @@
 //
 #include "ui/widgets/dropdown_menu.h"
 
+#include "ui/screen_reader_mode.h"
+
 #include <QtGui/QtEvents>
 
 namespace Ui {
@@ -31,6 +33,7 @@ DropdownMenu::DropdownMenu(QWidget *parent, const style::DropdownMenu &st) : Inn
 //}
 
 void DropdownMenu::init() {
+	InnerDropdown::setShownCallback([this] { showFinish(); });
 	InnerDropdown::setHiddenCallback([this] { hideFinish(); });
 
 	_menu->resizesFromInner(
@@ -228,6 +231,18 @@ void DropdownMenu::hideMenu(bool fast) {
 void DropdownMenu::childHiding(DropdownMenu *child) {
 	if (_activeSubmenu && _activeSubmenu == child) {
 		_activeSubmenu = SubmenuPointer();
+	}
+}
+
+void DropdownMenu::showFinish() {
+	// A popup menu opens for a screen reader as if from the keyboard: the
+	// first item is selected, takes the focus and is announced, and the
+	// arrows go on from there. A dropdown just appeared in the window,
+	// with nothing selected and nothing to hear. Do the same here, once
+	// the content is visible - a selected item that is still hidden
+	// behind the show animation could not hold the focus.
+	if (ScreenReaderModeActive()) {
+		_menu->setShowSource(TriggeredSource::Keyboard);
 	}
 }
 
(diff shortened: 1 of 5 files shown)


john-preston:
**Verdict: CHANGES_REQUIRED.** The mode gate and reuse of the existing menu selection machinery are appropriate. The new completion callback misses forced animation completion, and selection alone does not restore focus after a hide is reversed.

## Findings

### P2 — Complete the shown transition when an opening animation is forced to finish

The new [paint completion call](https://github.com/desktop-app/lib_ui/blob/9e2868abf545eebaa6b9c985949c1d269e660498/ui/widgets/inner_dropdown.cpp#L125) handles natural panel-animation completion. However, [finishAnimating](https://github.com/desktop-app/lib_ui/blob/9e2868abf545eebaa6b9c985949c1d269e660498/ui/widgets/inner_dropdown.cpp#L205) still stops the show animation, resets its data, and calls only `showChildren()`. With no opacity animation, no later callback remains to invoke `showFinished()`.

A deterministic source-level reproducer is a populated, unselected dropdown with the pointer outside it: call `showAnimated()`, then `finishAnimating()` before completion. The same path runs when a content-size change calls `resizeToContent()` during opening, for example adding an action that increases height. The content is visible but the new first-item selection, focus, and announcement never happen; focus remains on the opener. This is an uncovered completion path, not a claim that the normal attach menu always resizes while opening.

Route successful forced-show completion through the same once-per-transition hook. Keep hide cancellation distinct: `hideFast()` also calls `finishAnimating()`, so blindly changing every `showChildren()` to `showFinished()` would focus/announce a menu immediately before hiding it. Consolidate completion accounting with `showFast()` to avoid two calls.

### P2 — Reestablish focus and its return target when hiding is canceled

[showFinish](https://github.com/desktop-app/lib_ui/blob/9e2868abf545eebaa6b9c985949c1d269e660498/ui/widgets/dropdown_menu.cpp#L244) assumes selecting the first item also focuses it. That is false after a hide/reopen reversal. Start with a reader-mode dropdown showing and its first item selected/focused; start an animated hide, then call `showAnimated()` again before the fade completes. Auto-hiding dropdowns expose this through leaving and quickly reentering their toggle.

[Hide start](https://github.com/desktop-app/lib_ui/blob/9e2868abf545eebaa6b9c985949c1d269e660498/ui/widgets/inner_dropdown.cpp#L289) invokes `maybeReturnFocus()`, which consumes the saved opener and focuses it. Reversal neither saves that target again nor clears selection. On completion, [Menu::setSelected](https://github.com/desktop-app/lib_ui/blob/9e2868abf545eebaa6b9c985949c1d269e660498/ui/widgets/menu/menu.cpp#L452) returns because the first item is already selected, so the visible menu does not reacquire keyboard focus. If a later item was selected before hiding, completion changes to the first and does focus it, but the saved return target is empty; the next hide cannot explicitly restore the opener.

Treat selection and focus as separate responsibilities at reopening, and preserve or reacquire the focus-return target across a canceled hide. Retain the existing rule that dismissal must not steal focus back if the user has moved it outside the menu.

PR #364, opened 2026-09-04
Read the emoji of a field to a screen reader

rezabakhshilaktasaraei:
The emoji in the document of an `InputField` are inline objects, one U+FFFC (object replacement character) each. Qt's accessible interface for a `QTextEdit` returns the document text as is, so a screen reader reading the field got U+FFFC at an emoji and said nothing — the emoji could be copied (the text taken from the field resolves the objects), but was silent under the cursor and within a line.

Give the inner editor an accessible interface of its own from the factory: a `QAccessibleWidget` implementing `QAccessibleTextInterface` and `QAccessibleEditableTextInterface` on the public document API (`QTextDocument`, `QTextCursor`, `QTextLayout`, `QAccessible::qAccessibleTextBoundaryHelper`), that exposes the document text with the emoji's text (plain emoji from the image format, custom emoji from its stored text) in place of each object. The offsets of the interface are those of the exposed text — an emoji of two code units takes two offsets — mapped to document positions on the way in and back on the way out, so the caret, the selection and the word boundaries the platform derives from the text line up, and a range never cuts an emoji in two. Other objects (a collapsed quote) are left as they are.

No private Qt headers. Verified with NVDA on Windows: reading by character, word and line, caret and selection, deleting and inserting through the interface.


ilya-fedin:
> Uses `<private/qaccessiblewidgets_p.h>` (`Qt::WidgetsPrivate` is already linked).

How other Qt applications deal with that? I don't believe they use private API for a crucial thing such as accessibility. I'm currently working on private API less build since it private API use breaks badly on Linux (some distros could leave the user's system in a broken state due to tdesktop package requiring exact old Qt version due to private API while repos already providing a newer minor/patch Qt version).

rezabakhshilaktasaraei:
They implement `QAccessibleTextInterface` themselves — Qt's `QAccessibleTextEdit` is private exactly because applications are expected to provide their own interface for their own text widgets. Rewritten that way: the interface is now a `QAccessibleWidget` on the public document API, no private headers. While at it, the offsets are now those of the exposed text (an emoji of two code units takes two offsets), so word boundaries and selections line up with what the screen reader sees.

john-preston:
**CHANGES_REQUIRED.** Preserve the public-Qt approach, but correct text-change events, collapsed ranges, formatting attributes and horizontal scrolling. No ordinary keyboard/mouse regression was found in the changed paths.

1. **P2 — Translate native text-change notifications into the exposed coordinate space.** [Mapping implementation](https://github.com/desktop-app/lib_ui/blob/9909ccf20ed038d5f16ba6a84de65aa435daf569/ui/widgets/fields/input_field.cpp#L2039-L2047). Consider document `a/U+FFFC/b` exposed as `a❤️b`: the heart plus variation selector occupies two UTF-16 positions. Typing immediately before `b` generates native change position 2, while the exposed interface places it at 3. Removing the emoji reports one removed character rather than two. Nothing in this PR translates those events.

Configured Qt 5.15.19 and [Qt 6.11.2's producer](https://github.com/qt/qtbase/blob/v6.11.2/src/widgets/widgets/qwidgettextcontrol.cpp#L617-L639) use raw document offsets, selected text and removal lengths. Their [AT-SPI consumer](https://github.com/qt/qtbase/blob/v6.11.2/src/gui/accessible/linux/atspiadaptor.cpp#L1107-L1153) forwards change payloads directly, making incremental notifications inconsistent with queried text. **Caret is different:** both AT-SPI versions re-query `cursorPosition()`; selection notifications carry no range. Qt5 Windows UIA/Cocoa send payload-free notifications, so the author's successful NVDA run does not settle this defect. Provide one coherent event translation path, including pre-change expanded removal spans, and suppress/replace the contradictory original event rather than emitting a second event alongside it.

2. **P2 — Keep collapsed ranges collapsed.** [Endpoint rounding](https://github.com/desktop-app/lib_ui/blob/9909ccf20ed038d5f16ba6a84de65aa435daf569/ui/widgets/fields/input_field.cpp#L2053-L2064) maps starts down and ends up independently. In the same example, exposed `[2,2]` becomes document `[1,2]`. Thus `setSelection(0,2,2)` selects the emoji, `deleteText(2,2)` deletes it, and `replaceText(2,2,"X")` replaces it. An empty operation must not consume content. Qt5 UIA character-range movement increments integer endpoints, and its `Select()` forwards them through `addSelection()`, so interior endpoints are a real bridge input, not just malformed calls. Normalize a range jointly: preserve equal endpoints at one snapped position, keep empty reads/deletions empty, and expand only non-empty ranges, with consistent reverse-selection handling. This conclusion is a source/algorithm trace, not a runtime reproduction.

3. **P2 — Preserve formatting attributes and run boundaries.** [The new attributes method](https://github.com/desktop-app/lib_ui/blob/9909ccf20ed038d5f16ba6a84de65aa435daf569/ui/widgets/fields/input_field.cpp#L1941-L1950) returns no attributes and one field-wide range. Its “one font” rationale conflicts with [PrepareTagFormat](https://github.com/desktop-app/lib_ui/blob/9909ccf20ed038d5f16ba6a84de65aa435daf569/ui/widgets/fields/input_field.cpp#L1129-L1177), which applies bold, italic, underline and monospace to individual runs. Native QTextEdit accessibility exposes formatting; AT-SPI forwards it, and Cocoa uses the boundaries for style-range queries. A formatted message made entirely of ordinary letters therefore loses existing accessibility information. Derive the relevant attributes from public QTextCharFormat/QTextBlockFormat data and translate each run's endpoints.

4. **P2 — Restore horizontal substring scrolling.** [scrollToSubstring](https://github.com/desktop-app/lib_ui/blob/9909ccf20ed038d5f16ba6a84de65aa435daf569/ui/widgets/fields/input_field.cpp#L1927-L1939) adjusts only the vertical scrollbar. [SingleLine fields use NoWrap](https://github.com/desktop-app/lib_ui/blob/9909ccf20ed038d5f16ba6a84de65aa435daf569/ui/widgets/fields/input_field.cpp#L2308-L2312); a long plain-text value can have an offscreen prefix/suffix. An accessibility request to reveal it now leaves it hidden. Native QTextEdit's ensure-visible path adjusts both axes. Implement both axes through public geometry/scrollbar APIs, including RTL horizontal offsets, without moving the user's caret or selection.

5. **P3 — Keep the local class declaration separate from its implementation.** [The local class](https://github.com/desktop-app/lib_ui/blob/9909ccf20ed038d5f16ba6a84de65aa435daf569/ui/widgets/fields/input_field.cpp#L1734-L1743) defines substantial methods inline despite REVIEW.md's declaration-only class rule. Place definitions after the compact declaration and sort touched includes. This code-structure finding is separate from the four functional blockers.

PR #366, opened 2026-09-08
Animate a wheel notch instead of jumping to its target.

abdallah-shehawey:
A wheel notch moves a scroll straight to where it is going, so reading a long
list with the wheel is a series of jumps with nothing to follow between them.
Each notch now eases to that same position over 260ms, so the place being read
stays visible while it travels.

### Only a classic notch is animated

A trackpad, a touch screen and the momentum tail after them already arrive as a
smooth stream of phased deltas, and animating on top of one would only make it
worse — so only a `Qt::NoScrollPhase` event with a full `|angleDelta().y()| >= 120`
that dominates its horizontal component is taken. The modifier wheels that page a
scroll are left alone, and the top and bottom edges stay with the scroll, where the elastic overscroll is stretched and where a
list is asked for more content to append below the last item.

The delta the animation travels is the one `QAbstractSlider` would have applied at
once — `wheelScrollLines() * singleStep()`, clamped to `pageStep()` — so taking the
event over changes how the scroll travels, not where it arrives.

### Nothing above the library opts in or out

This is the point of the change. It started as telegramdesktop/tdesktop#31258,
where it needed 10 files of application code — an install call per scroll and an
opt-out at every scroll that drives the wheel itself. @ilya-fedin asked for a
reusable component in lib_ui instead, shaped like `KineticScroller`: a wheel event
transformation that needs no changes in the application repo. With this branch,
tdesktop is byte-identical to `dev`.

**`ScrollArea`** takes the notch from an event filter installed together with its
scroll bar. Qt hands a wheel event to the scroll bar, and filters are activated in
reverse order of installation — nobody can filter a scroll before it exists, so
every handler an owner installs later is offered the notch first and the animation
only ever takes what is left. `Info::FlexibleScrollHelper`, which filters that same
bar to snap a pinned header, keeps working untouched.

**`ElasticScroll`** takes it in the `Qt::NoScrollPhase` branch, where it would have
applied the delta at once, after its own `_customWheelProcess` has declined.

`willScrollTo()` gets the two bounds it was already clamping to named as
`scrollToMin()` / `scrollToMax()`, so the animation can ask for the same ones.

### Checked

- Qt delivers a wheel event to the vertical scroll bar **even though lib_ui hides
  it** with `Qt::ScrollBarAlwaysOff` — confirmed with a standalone Qt 6.11 program,
  which is what makes the scroll-bar filter a valid place to hook.
- A filter installed later is activated first and can consume the event before the
  one installed with the bar ever sees it — confirmed in the same rig, which is what
  keeps every application handler ahead of the animation.
- For a standard notch the helper reproduces Qt's own movement exactly (60px at
  `wheelScrollLines() = 3`, `singleStep() = 20`).
- `lib_ui` and a full `Telegram` build are clean on Linux/GCC, no new warnings.


ilya-fedin:
> `anim::Disabled()` is respected

Are you sure it's a right thing to do? This makes smooth scroll get disabled on battery

23rd:
Remove all comments.

abdallah-shehawey:
@ilya-fedin You're right, dropped it — the check is gone from the pushed branch.

It was inconsistent anyway: no other scroll motion in these two files consults `anim::Disabled()`. The touch flick in `ScrollArea`, `ElasticScroll`'s overscroll return and `KineticScroller` all keep animating with animations off, so a wheel notch stopping to jump while a flick still glides was the odd one out. And since it is `PowerSaving::kAnimations` that drives the flag, with force-on-battery it would have changed how the wheel behaves the moment the laptop is unplugged.

@23rd Done, all comments removed.

Both in the force-push above, rebased on current master.

PR #367, opened 2026-09-13
Keep fullscreen panel buttons on their own Space on macOS

meh7an:
In fullscreen, `SeparatePanel` shows its close, back and menu buttons as separate frameless `Qt::Tool` windows above the panel. On macOS Qt creates such windows as panels with `NSWindowCollectionBehaviorMoveToActiveSpace`, and at the floating window level they are not managed by Spaces, so they follow whichever Space is active.

A panel in native fullscreen has a Space of its own. Switching to another desktop brought the buttons along, so they showed up over the other desktop, and after the app lost focus they could stay there instead of returning to the panel. For example, a mini-app opened in fullscreen showed its close and menu buttons on the main desktop.

`Platform::KeepOnCurrentSpace()` removes `MoveToActiveSpace` from a window and marks it as managed, so it stays on the Space it was shown on. `SeparatePanel::initFullScreenButton()` calls it right after showing each button, and the buttons of a fullscreen panel are created on its Space. It only reads the existing native window instead of calling `winId()`, which would make the panel's children native and could put them above an embedded web view. On other platforms it does nothing.

**Testing**

Built into Telegram Desktop and tested on macOS (Debug), with a mini-app in fullscreen:

- Switching to another desktop and back, also after activating another app, keeps the buttons only on the fullscreen panel's Space.
- The mini-app's web content is shown every time it is opened in fullscreen.


ui/platform/mac/ui_utility_mac.mm
```diff
@@ -107,6 +107,17 @@ void AcceptAllMouseInput(not_null<QWidget*> widget) {
 	[window setIgnoresMouseEvents:NO];
 }
 
+void KeepOnCurrentSpace(not_null<QWidget*> widget) {
+	const auto view = reinterpret_cast<NSView*>(widget->internalWinId());
+	const auto window = view ? [view window] : nil;
+	if (!window) {
+		return;
+	}
```
ilya-fedin: Follow the existting pattern

```suggestion
	NSWindow *window = [reinterpret_cast<NSView*>(widget->winId()) window];
```

ilya-fedin:
please squash the commits

PR #368, opened 2026-09-14
Fix visual glitch when moving maximized window across monitors on Windows

PolimerCode:
### Problem
On Windows 11, moving a maximized window to another monitor using the \Win + Shift + Left / Right\ shortcut causes a visual glitch: an unpainted ~16px strip appears along the right edge showing a duplicated ghost scrollbar, and the bottom border is slightly cut off.

### Root Cause
On Windows 11 (\
ativeResize() == true\), \margins\ is already initialized with zero horizontal and bottom margins (\QMargins(0, r.top, 0, 0)\) because DWM handles borders natively.
The legacy \WS_MAXIMIZE\ / \_marginsDelta\ block in \WindowHelper::updateMargins()\ subtracted \_marginsDelta\ (-8px border offset) from \margins\, resulting in +8px custom margins on the sides and bottom. Qt then shrank the client widget by 16px horizontally, leaving an unpainted 16px strip where DWM retained stale bits.

### Changes
Guard the legacy \WS_MAXIMIZE\ and \_marginsDelta\ handling in \WindowHelper::updateMargins()\ with \!nativeResize()\.
On Windows 11, margins depend solely on style and DPI, avoiding obsolete monitor work area adjustments and eliminating redundant listeners.


ui/platform/win/ui_window_win.cpp
```diff
@@ -891,25 +926,29 @@ void WindowHelper::updateMargins() {
 			m.right - w.right,
 			m.bottom - w.bottom);
 
-		margins.setLeft(margins.left() - _marginsDelta.left());
-		margins.setRight(margins.right() - _marginsDelta.right());
-		margins.setBottom(margins.bottom() - _marginsDelta.bottom());
+		if (!nativeResize()) {
```
ilya-fedin: this looks wrong, `nativeResize()` is already accounted in the `margins` variable value

ui/platform/win/ui_window_win.cpp
```diff
@@ -891,25 +926,29 @@ void WindowHelper::updateMargins() {
...
-			0,
-			0,
-			w.right - w.left - _marginsDelta.left() - _marginsDelta.right(),
-			w.bottom - w.top - _marginsDelta.top() - _marginsDelta.bottom(),
-			(SWP_NOMOVE
-				| SWP_NOSENDCHANGING
-				| SWP_NOZORDER
-				| SWP_NOACTIVATE
-				| SWP_NOREPOSITION));
+		if (!nativeResize()) {
```
ilya-fedin: Is it right given that `_marginsDelta` accounts `nativeResize()` in its value?

ui/platform/win/ui_window_win.cpp
```diff
@@ -625,6 +647,17 @@ bool WindowHelper::filterNativeEvent(
 		if (_shadow) {
 			_shadow->update(WindowShadow::Change::Moved);
 		}
+		const auto hMonitor = MonitorFromWindow(
+			_handle,
+			MONITOR_DEFAULTTONEAREST);
+		if (_currentMonitor != hMonitor) {
+			_currentMonitor = hMonitor;
+			if (GetDpiForWindowSupported()) {
+				_dpi = GetDpiForWindow(_handle);
```
ilya-fedin: Why should we handle dpi change here given that Windows sends a WM_DPICHANGED?

ui/platform/win/ui_window_win.cpp
```diff
@@ -436,6 +439,25 @@ void WindowHelper::init() {
...
+			&QWindow::screenChanged,
+			[=](QScreen *screen) {
+				if (!_handle) {
+					return;
+				}
+				_currentMonitor = MonitorFromWindow(
+					_handle,
+					MONITOR_DEFAULTTONEAREST);
+				if (GetDpiForWindowSupported()) {
+					_dpi = GetDpiForWindow(_handle);
```
ilya-fedin: Why should we handle dpi change here given that Windows sends a WM_DPICHANGED?

ui/platform/win/ui_window_win.cpp
```diff
@@ -436,6 +439,25 @@ void WindowHelper::init() {
 		&QWindow::windowStateChanged,
 		handleStateChanged);
 
+	if (const auto handle = window()->windowHandle()) {
```
ilya-fedin: windowHandle just was used without a nullptr check multiple lines above

ui/platform/win/ui_window_win.cpp
```diff
@@ -436,6 +439,25 @@ void WindowHelper::init() {
...
+		Ui::Connect(
+			handle,
+			&QWindow::screenChanged,
```
ilya-fedin: QWindow could be recreated during widget lifetime, it's better to use `window()->screenValue()`

ui/platform/win/ui_window_win.cpp
```diff
@@ -436,6 +439,25 @@ void WindowHelper::init() {
...
+			[=](QScreen *screen) {
+				if (!_handle) {
+					return;
+				}
+				_currentMonitor = MonitorFromWindow(
```
ilya-fedin: Why do we need to handle Qt's screen change event together with WM_MOVE that includes window moves between screens?

ui/platform/win/ui_window_win.cpp
```diff
@@ -877,7 +910,9 @@ void WindowHelper::updateMargins() {
 		GetWindowRect(_handle , &w);
 		m = w;
 
-		HMONITOR hMonitor = MonitorFromRect(&w, MONITOR_DEFAULTTONEAREST);
+		HMONITOR hMonitor = MonitorFromWindow(
```
ilya-fedin: It should be the same given that the rect was just retrieved?

ilya-fedin:
> ###
> 
>     1. Guard horizontal and bottom margin adjustments with `!nativeResize()`.
> 
>     2. Use `MonitorFromWindow` instead of `MonitorFromRect` in `updateMargins()` to avoid reading stale transition coordinates.
> 
>     3. Handle monitor change in `WM_MOVE` to ensure DPI and margins are refreshed and the window is repainted.
> 
>     4. Safely connect `QWindow::screenChanged` only when `window()->windowHandle()` is valid.

These changes are very questionable, it feels like what really fixed it could be just one change of all that and everything else is unneeded. e.g. I see that you use `Ui::ForceFullRepaint(window());` in your code on DPI changes, maybe the only thing needed is adding it to the existing `_dpi.value()` handler?

PolimerCode:
Thanks for the review! Regarding the `_dpi.value()` suggestion:
This issue reproduces on multi-monitor setups where both monitors have the exact same dpi (for example, the test setup here is two 100% scale displays: 1080p and 1440p). When moving a maximized window via `Win + Shift + Left/Right` between monitors with identical scale, `_dpi` does not change, so `_dpi.value()` is never triggered.
The root cause of the 16px duplicate strip is twofold:
1. On same-dpi monitors, `updateMargins()` was never called upon moving to the other screen, leaving margins un-updated for the new screen's work area.
2. In `updateMargins()`, on windows 11 (`nativeResize() == true`), subtracting `_marginsDelta` (`-8px`) caused positive custom margins (`+8px` left, right, bottom). Qt then shrank the client widget by 16px horizontally and 8px vertically, leaving an unpainted 16px strip on the right where DWM retained stale bits.

You are right that having both `WM_MOVE` monitor tracking and `QWindow::screenChanged` is redundant. We can streamline the patch and drop the `WM_MOVE` handling and `_currentMonitor` member, keeping only a clean `screenChanged` connection to call `updateMargins()` and `Ui::ForceFullRepaint()`.
Would you like me to update the PR to this minimal version?

ilya-fedin:
> When moving a maximized window via `Win + Shift + Left/Right` between monitors with identical scale, `_dpi` does not change, so `_dpi.value()` is never triggered.

If DPI is unrelated, why you're touching it?

> 2\. In `updateMargins()`, on windows 11 (`nativeResize() == true`), subtracting `_marginsDelta` (`-8px`) caused positive custom margins (`+8px` left, right, bottom). Qt then shrank the client widget by 16px horizontally and 8px vertically, leaving an unpainted 16px strip on the right where DWM retained stale bits.

Looks like you have some issue with triaging. It substracts only the top margin with `nativeResize() == true`. Maybe something somewhere is overlooked but generally you don't need to completely disabled it, `nativeResize()` is already accounted in the margins.

ilya-fedin:
Maybe actual problem is that updateMargins uses rcWork instead of the same logic as WM_NCCALCSIZE uses?

PolimerCode:
Exactly, `WM_NCCALCSIZE` already trims `r` by `borderWidth` (`SM_CXSIZEFRAME + SM_CXPADDEDBORDER` = 8px at 100% scale):
```
r->left += borderWidth;
r->right -= borderWidth;
r->bottom -= borderHeight;
```
So windows DWM already sets the client area to match the monitor work area
Then in `updateMargins()`:
`_marginsDelta` computes `w.left - mi.rcWork.left` (= -8px on each side).
Because `nativeResize()` initialized `margins` to `QMargins(0, r.top, 0, 0)` (0 on the sides), doing
```
margins.setLeft(margins.left() - _marginsDelta.left()); // 0 - (-8) = +8
margins.setRight(margins.right() - _marginsDelta.right()); // 0 - (-8) = +8
margins.setBottom(margins.bottom() - _marginsDelta.bottom()); // 0 - (-8) = +8
```
adds an extra 8px to `margins.left`, `margins.right`, and `margins.bottom`. Qt then subtracts these margins from the widget, shrinking it by an additional 16px horizontally (8 + 8). This leaves the exact 16px unpainted strip along the right edge where DWM retains stale bits.

So `WM_NCCALCSIZE` and `updateMargins()` are indeed doublecounting the border on windows 11.
How would you prefer to align them?
Skip subtracting `_marginsDelta` for left/right/bottom when `nativeResize()` is true, since `WM_NCCALCSIZE` already accounted for them.
Or align `updateMargins()` to use the same `borderWidth` / `borderHeight` calculation directly.
And we can connect `QWindow::screenChanged` to `updateMargins()` so margins refresh when moving between monitors

ilya-fedin:
> `_marginsDelta` computes `w.left - mi.rcWork.left` (= -8px on each side).

How did you count that? Window rect should be exactly the size of monitor so it should always be 0. I'm slowly remembering what this code is doing and looks like what is doing is exactly querying WM_NCCALCSIZE via GetWindowRect and calculating the difference between the actual size (from WM_NCCALCSIZE/GetWindowRect) and the monitor resolution. So, it seems, if everything works right, _marginsDelta should be always zero? Maybe it's just a leftover from older code and both WS_MAXIMIZE/_marginsDelta codepaths aren't needed nowadays :thinking: 

> And we can connect `QWindow::screenChanged` to `updateMargins()` so margins refresh when moving between monitors

Why should that be done? If DPI has changed, it already calls updateMargins, if it didn't change, there's no reason to update margins on screen change.

PolimerCode:
> How did you count that? Window rect should be exactly the size of monitor so it should always be 0.

In Win32, when a window with `WS_THICKFRAME` is maximized, windows intentionally expands `GetWindowRect` beyond the monitor's work area by `SM_CXSIZEFRAME + SM_CXPADDEDBORDER` (8px on 100% scale) to push the resize borders off-screen
So for a maximized window, `GetWindowRect` on a 1080p monitor returns `left = -8, top = -8, right = 1928, bottom = 1048`.
Comparing `w` to `mi.rcWork` (`0, 0, 1920, 1040`) yields `w.left - mi.rcWork.left == -8`.

> So, it seems, if everything works right, _marginsDelta should be always zero? Maybe it's just a leftover from older code and both WS_MAXIMIZE/_marginsDelta codepaths aren't needed nowadays :thinking:

Your intuition here actually is spot on. Looking at commit `fd55e9b7` ("Adapt WindowHelper::updateMargins for windows 11 native resize"), `updateMargins()` originally had an early return on windows 11. When it was enabled for windows 11 with `margins = QMargins(0, r.top, 0, 0)`, the `WS_MAXIMIZE` block below it was kept, which subtracted `_marginsDelta` (`-8px`) from `0`, turning the horizontal margins into `+8px`.
On Windows 11, DWM handles maximized borders natively, so the `WS_MAXIMIZE` / `_marginsDelta` adjustment in `updateMargins()` is indeed obsolete and harmful

> If DPI has changed, it already calls updateMargins, if it didn't change, there's no reason to update margins on screen change.

You are right. Once `WS_MAXIMIZE` / `_marginsDelta` is bypassed on Windows 11, `margins` depends solely on DPI and window style, not on monitor work area coordinates. That means we don't need `screenChanged`, `WM_MOVE`, or `_currentMonitor` at all
The entire fix simply becomes guarding the `WS_MAXIMIZE` / `_marginsDelta` block with `!nativeResize()`.
I'll update the PR to this minimal change.

ilya-fedin:
Actually I misread GetWindowPos as GetClientPos. idk which AI you're using (your comments look AI generated, sorry if you don't) but seems garbage, it agrees with everything you feed it with even if that's false. Try this patch:
```
diff --git a/ui/platform/win/ui_window_win.cpp b/ui/platform/win/ui_window_win.cpp
index f34c1df2..0264238d 100644
--- a/ui/platform/win/ui_window_win.cpp
+++ b/ui/platform/win/ui_window_win.cpp
@@ -872,6 +872,7 @@ void WindowHelper::updateMargins() {
 	auto margins = nativeResize()
 		? QMargins(0, r.top, 0, 0)
 		: QMargins(r.left, r.top, -r.right, -r.bottom);
+	_rcWorkDelta = QMargins();
 	if (style & WS_MAXIMIZE) {
 		RECT w, m;
 		GetWindowRect(_handle , &w);
@@ -885,16 +886,15 @@ void WindowHelper::updateMargins() {
 			m = mi.rcWork;
 		}
 
-		_marginsDelta = QMargins(
+		_rcWorkDelta = QMargins(
 			w.left - m.left,
 			w.top - m.top,
 			m.right - w.right,
 			m.bottom - w.bottom);
 
-		margins.setLeft(margins.left() - _marginsDelta.left());
-		margins.setRight(margins.right() - _marginsDelta.right());
-		margins.setBottom(margins.bottom() - _marginsDelta.bottom());
-		margins.setTop(margins.top() - _marginsDelta.top());
+		const auto normal = margins;
+		margins = QMargins(r.left, r.top, -r.right, -r.bottom) - _rcWorkDelta;
+		_marginsDelta = normal - margins;
 	} else if (!_marginsDelta.isNull()) {
 		RECT w;
 		GetWindowRect(_handle, &w);
@@ -945,8 +945,8 @@ void WindowHelper::fixMaximizedWindow() {
 				0,
 				0,
 				0,
-				m.right - m.left - _marginsDelta.left() - _marginsDelta.right(),
-				m.bottom - m.top - _marginsDelta.top() - _marginsDelta.bottom(),
+				m.right - m.left - _rcWorkDelta.left() - _rcWorkDelta.right(),
+				m.bottom - m.top - _rcWorkDelta.top() - _rcWorkDelta.bottom(),
 				SWP_NOMOVE | SWP_NOSENDCHANGING | SWP_NOZORDER | SWP_NOACTIVATE | SWP_NOREPOSITION);
 		}
 	}
diff --git a/ui/platform/win/ui_window_win.h b/ui/platform/win/ui_window_win.h
index acab8023..3ec09dc3 100644
--- a/ui/platform/win/ui_window_win.h
+++ b/ui/platform/win/ui_window_win.h
@@ -80,7 +80,7 @@ private:
 	rpl::event_stream<HitTestResult> _systemButtonDown;
 	std::optional<WindowShadow> _shadow;
 	rpl::variable<uint> _dpi;
-	QMargins _marginsDelta;
+	QMargins _marginsDelta, _rcWorkDelta;
 	HWND _handle = nullptr;
 	bool _updatingMargins = false;
 	bool _isFullScreen = false;
```

PolimerCode:
Applied your patch and updated the branch, thanks

ilya-fedin:
But did you check whether it actually helps?

PolimerCode:
Just made a build and verified. Moving windows across displays correctly now.

ilya-fedin:
Nice, thanks

john-preston:
Thanks!

